PR #23, opened 2014-12-06
added option to show function arguments

zer0def:
Figured I'd give a hand, since I find your package immensely helpful in debugging. :)


joerick:
I'm closing this out because of inactivity, and because there have been too many changes in master for this to still work. In general there could be a good idea here, but I think the approach needs consideration of how this data would **a)** be treated by the processors (especially the time-aggregator) and **b)** be rendered. I have a couple of ideas about this so let me know if it's still interesting and we can talk more!

PR #24, opened 2015-01-21
Add a flamegraph charting output

cpennington:
This adds a flame charting method (as described here: https://github.com/brendangregg/FlameGraph)


joerick:
Hey, thanks for the PR. I just downloaded and run it, but I'm a bit confused about the output.

image

This is what I get when I run it on the django_test project. Maybe it's not a good example.

Perhaps you have an example or screenshot that shows a better use-case?

cpennington:
Ah, interesting. I actually hadn't tried it using the django integration, so I'm not 100% sure how well that'll work.

Here's an example of a test run from an issue I was debugging:

flamegraph

You can see that `test_generate_find_timings` was the main test (as it takes up 100% of the width) and then that xml parsing took up noticable time (the `import_from_xml -> _import_course_asset_metadata -> from_xml` chain). You can hover over any elements to get more detailed information (which file they come from, how much absolute time they take).

I actually think that some of the details on the display could be improved. The highest level stack frame doesn't get a full frame height, for some reason, and I think I left some debugging info in the text popup.

Seems like a legend might also be useful, to help interpretation.

cpennington:
Also, it would be nice to give a better indication of what's going on when you've zoomed into the graph (a more gradual transition, and dimming of the stack above your zoom level, for instance).

joerick:
Hi @cpennington 

Just to keep you updated- I really want to get flamegraphs into pyinstrument, and I think your method of rendering using SVG is great.

I'm actually working on some refactoring so we can put in time-based profiling, so the chart actually has time running along the x-axis. You might have noticed this already, but because the chart above is made using a call tree that's aggregated and sorted, the longest-running of the child frames always comes first, and repeats are not represented anywhere.

I should have a branch you can pull from soon and then we can get this thing moving! :)

cpennington:
Oh, that's cool. I think both the time-based and the aggregate versions would be interesting for different purposes, and it shouldn't be hard to adapt the d3 code I have to handle either one.

joerick:
I've just pushed the refactorings to my develop branch. You can merge, or if you like, you could add me to the collaborators of your fork and I'll do the merge and push it back up to this PR.

I've added stubs for the code to go into a FlameRenderer object in renderers.py.

cpennington:
I rebased this on top of develop, but this PR is against master. I'm going to open a new one against develop.

cpennington:
Closed in favor ef #25.

PR #27, opened 2016-02-06
context manager & ipython magic

maxim-lian:
Two features here - ipython magic, and a contextmanager. Let me know your thoughts. It probably needs some runtime checks to ensure it's being run in ipython etc

jupyter / ipython notebook usage:

``` python
import pyinstrument
%iprun long_running_calc()
```

context manager usage:

```
from pyinstrument import instrument_profile
with instrument_profile():
    long_running_calc()
```

### DIFF
--- a/pyinstrument/__init__.py
+++ b/pyinstrument/__init__.py
@@ -1,1 +1,7 @@
-from pyinstrument.profiler import Profiler
+from pyinstrument.profiler import Profiler, instrument_profile, PyInstrumentMagic
+
+
+# TODO: ensure this doesn't break if ipython isn't available
+from IPython import get_ipython
+ip = get_ipython()
+ip.register_magics(PyInstrumentMagic)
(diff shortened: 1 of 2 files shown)


MaximilianR:
@superbobry Thanks for the comments. 

Are you involved with the project - i.e. if I make those changes will this get merged?

superbobry:
Nope, I'm just a user. Maybe @joerick can comment?

joerick:
Hi guys, thanks for this. I'm not a big ipython user but can see how this could be helpful, for sure. I'd like to merge this if we fix the import stuff as @superbobry mentioned, and move all the ipython code into a separate file ipython.py or similar, so it's clear that it's extending on functionality in the core of pyinstrument.

p.s. thanks for the review @superbobry!

joerick:
Closing due to inactivity... unless you still have interest in this @maxim-lian ?

maxim-lian:
Ah, sorry I let this drop! Particularly given it was close to the line. 

I'm not doing much profiling at the moment. That said, if there are people who would find this helpful, comment here and I'll commit to push this over the line if there's a few. 

Otherwise I may anyway when I'm profiling more

PR #29, opened 2017-10-09
Make Pyinstrument easier to use with new profiling mode, and a big refactor

joerick:
This PR adds a new profiling mode, that's as quick as signal-mode, but without the downsides. The setprofile mode has been implemented in a C extension (see joerick/pyinstrument_cext) as `setstatprofile()`.

The internals have also been refactored to split responsibilities more cleanly. Hopefully this will make the creation of new renderers (such as a flame-chart renderer) easier.

I've added a couple tests (that are checked in CI by Travis).

Still some work to clean up here, but I'm excited about this new version!


joerick:
The C extension (pyinstrument_cext) has [wheels](https://pypi.org/project/pyinstrument_cext/#files) for most packages (using [cibuildwheel](/joerick/cibuildwheel) :)). Does anyone think I need to provide a pure-python fallback (that will have a larger overhead and therefore might distort timings) for the few users that maybe can't use wheels or compile it from source for some reason?

asmeurer:
That's exciting. Are you building the wheels for all three platforms?

I guess it's up to you for the pure python, if you want to maintain the code. Even the standard library has a pure python version of the profile module (though I was never really clear why).

joerick:
> I guess it's up to you for the pure python, if you want to maintain the code

We can think about providing a pure python alternative if we get some people complaining in the issues :)

> Are you building the wheels for all three platforms?

Yep, using a tool I made called cibuildwheel. It's pure C with no additional dependencies so pretty easy to build.

joerick:
Here's the changelog entry for the curious:

> ### What's new in v1.0 ###
>
> -   Pyinstrument refactored to use a new profiling mode. Rather than using
>     signals, pyintrument uses a new statistical profiler built on
>     PyEval_SetProfile. This means no more main thread restriction, no more 
>     IO errors when using Pyinstrument, and no more 'setprofile' mode!
> 
> -   Renderers. Users can customize Pyinstrument to use alternative renderers
>     with the `renderer` argument on `Profiler.output()`, or using the `--renderer`
>     argument on the command line.
> 
> -   Recorders. To support other use cases of Pyinstrument (e.g. flame charts),
>     pyinstrument now has a 'timeline' recorder mode. This mode records captured
>     frames in a linear way, so the program execution can be viewed on a
>     timeline.

joerick:
This is now available in beta! Try it out with `pip install pyinstrument==2.0.0b1`, and please let me know if there are any issues for you!

Issue #15, opened 2014-06-19
Allow using `Profiler` as context manager


cool-RR:

MaximilianR:
+1

mjpieters:
I created a PR before I found the issues list, see #36.

joerick:
Fixed in #36

PR #36, opened 2018-03-09
Make Profiler a context manager

mjpieters:
Rather than have to explicitly start and stop, use Profiler as a context manager to auto-start and stop. Test included.

### DIFF
--- a/pyinstrument/profiler.py
+++ b/pyinstrument/profiler.py
@@ -37,6 +37,13 @@ class Profiler(object):
     def stop(self):
         setstatprofile(None)
 
+    def __enter__(self):
+        self.start()
+        return self
+
+    def __exit__(self, *args):
+        self.stop()
+
     def _profile(self, frame, event, arg):
         now = timer()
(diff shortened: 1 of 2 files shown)


mjpieters:
Just found issue #15, this PR would satisfy that request.

joerick:
Thanks for this mjpeiters! Let me have a think, I'm wondering if it would be useful to specify the output arguments in the `with` statement, so the user doesn't even need to capture the profiler object in a variable. What do you think?

mjpieters:
You don't *have* to capture the profiler with the `as` keyword, that's optional, just like with the file objects. You can also use:

    profiler = Profiler()
    with profiler:
        # profiled code

I don't think it makes sense at all to specify the output options the `with` statement. You may want to profile separate sections, and using the profiler as a CM should just echo the way you use `.start()` and `.stop()`.

joerick:
Fair enough. I like your second code sample the best - makes `profiler` look relevant outside the block. I think this way works best with the API as it stands :)

Thanks very much for the contribution!

PR #38, opened 2018-04-25
Enable Django versions to work without MiddlewareMixin.

z2kviswanathan:
Hi,

Some of the Django versions don't ship this class. We just decided to get over this problem like so:


joerick:
Neat trick! Which Django versions does this add support for?

z2kviswanathan:
For versions prior to the 1.10.x series; For us, we were running the 1.7.x series. We could run it with this change.

joerick:
Nice one! Not at a computer so can't cut a release now but feel free to bug me in a few days if I haven't got around to it :)

Issue #43, opened 2018-07-17
Running modules


iddan:
I'd like to trace a program requiring execution with `python -m`. How should I do it? I know Python 3.7 has introduced support for it in some of the script executing modules.

joerick:
Here's the commit that added it to cProfile: https://github.com/python/cpython/commit/7973e279a21999f134aff92dd6d344ec4591fae9

Nothing clever, just adding another option to the command line parser. PRs welcome!

PR #45, opened 2018-07-30
Support running modules

iddan:
Solves #43. It was a pleasure to get into the codebase. Would suggest adding setting up a venv into the readme under "contributing" or something.

### DIFF
--- a/pyinstrument/__main__.py
+++ b/pyinstrument/__main__.py
@@ -2,9 +2,9 @@ from optparse import OptionParser
 import sys
 import os
 import codecs
+import runpy
 from pyinstrument import Profiler
 from pyinstrument.profiler import get_recorder_class, get_renderer_class
-from importlib.util import find_spec
 from .six import exec_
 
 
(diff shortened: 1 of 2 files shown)


iddan:
@joerick can you please review it?

joerick:
sure thing. thanks for the contribution! i'll get around to it, just busy at work this week. I'll see if i can take a look at the weekend. thanks!

pyinstrument/__main__.py
```diff
@@ -44,64 +45,74 @@ def main():
...
+    sys.argv[:] = args
 
-        if options.output_renderer:
-            renderer = options.output_renderer
-        elif options.output_html:
-            renderer = 'html'
-        else:
-            renderer = 'text'
+    if options.module_name is not None:
+        progname = find_spec(options.module_name).origin
```
joerick: I'm not sure this is right - when I run `find_spec('pyinstrument').origin` in a shell it gives me the `__init__.py` file. It should be running the `__main__.py` file as I understand it.

The [cProfile module](https://github.com/python/cpython/blob/master/Lib/cProfile.py#L157) uses [runpy](https://docs.python.org/2.7/library/runpy.html) for this - it's probably best to copy that.

iddan:
Changed to runpy and made test succeed

iddan:
If you can please post when it is released to PyPI.

joerick:
Sorry for the delay - just addressing a bug I found before releasing... I had to make a minor change to this for programs that use ArgumentParser to work. Most programs assume sys.argv has at least one entry. see 2c0599063ba7ac026aba74249cc0e7ec697f5a6c

joerick:
released as 2.1.1!

iddan:
Maybe update should be state in README?

PR #49, opened 2018-09-15
[WIP] Version 3.0 - hide libraries and UI refresh

joerick:
Hey - just wanted to share some WIP with you all on some stuff I'm working on for pyinstrument.

The main feature is the ability to hide traces through libraries that you're using. So instead of showing you loads of frames going through the internals of e.g. urllib, it lets you focus on your code. Here's an example:

| Before | After |
| --- | ---
| image | image | 

A few other things to note-

- 'Entry' frames of hidden groups are shown, so you know which call is the problem
- Really slow frames in the groups are shown too, e.g. the 'read' call on the socket
- Application code is highlighted in the console
- Additional metrics are shown at the top of the trace - timestamp, number of samples, duration, CPU time
- Hidden code is controlled by the `--hide` or `--hide-regex` options - matching on the path of the code files. 
  ```
    --hide=EXPR           glob-style pattern matching the file paths whose
                          frames to hide. Defaults to '*/lib/*'.
    --hide-regex=REGEX    regex matching the file paths whose frames to hide.
                          Useful if --hide doesn't give enough control.
  ```

- Outputting a timeline is supported from the command line.

  ```
    -t, --timeline        render as a timeline - preserve ordering and don't
                          condense repeated calls
  ```

- Because there are a few rendering options now, you can load a previous profiling session using `--load-prev` - pyinstrument keeps the last 10 sessions.

- Hidden groups can also call back into application code, that looks like this:
    
    image

Because this is a pretty big change internally, I've had to deprecate some APIs. I'm not sure how much they are used, but I think it's worth a version bump anyway so I'm tempted to call this version 3.0!

Still some things to do:

- [x] Remove pyinstrument internal frames from command line traces (and/or put the root at the caller of profiler.start()
- [x] HTML hide group support 
  - allow interaction - click the group to open it
  - visual refresh ([it might look like this](https://user-images.githubusercontent.com/1244307/45589750-2e80ff80-b923-11e8-83c5-ed19a7bc77d2.png))
- [x] timelines still aren't \_quite\_ right - they don't track the linearity of self-time. This would allow for the creation of super-accurate flame charts. Tempted to get this in to v3.0, since it might be a pretty big change also!
- [x] Tests for each processor

In the meantime, would love to hear feedback, ideas, thoughts, critiques!


joerick:
hello @iddan - quick question - to support the HTML renderer i've made a breaking change to the JSON output schema, now the root frame is at a property `root_frame` on the JSON output. would you be willing to change your online tool to look at `root_frame` and then fallback to the current behaviour if it's not found? I think your tool is the only consumer of the JSON at the moment so I'm thinking this breaking change will be simpler than doing two different JSON schemas within pyinstrument. any thoughts?

iddan:
Yeah, I'll change it. But I don't think backwards compatibility is crucial as it is easy to upgrade and there is no a lot of users currently

joerick:
awesome, thanks @iddan !

joerick:
If anyone is interested in giving it a test, a prerelease is now available on PyPI - `pip install pyinstrument --pre` for the latest. I'd love some bug reports / feedback!

PR #58, opened 2019-01-15
Provide additional arguments in output_text method.

sandrohuber:
Hi,

We use the python profiler in a gunicorn-driven server. It would be nice to still have the possibility to choose between verbose or (the new) compact profiling result text. In case of slow performance we often want to look into the performance of dependencies, because we might have configured them wrong.


joerick:
Thank you @sandrohuber !

PR #63, opened 2019-04-09
profiler: remove useless timer check

jd:
The timer check in Profiler.profile ought to be useless since the timing is
already handled by pyinstrument_cext.
In practice, except for rounding errors, time_since_last_profile is always >=
self.interval.

### DIFF
--- a/pyinstrument/profiler.py
+++ b/pyinstrument/profiler.py
@@ -70,9 +70,6 @@ class Profiler(object):
         now = timer()
         time_since_last_profile = now - self.last_profile_time
 
-        if time_since_last_profile < self.interval:
-            return
-
         if event == 'call':
             frame = frame.f_back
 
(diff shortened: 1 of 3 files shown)


jd:
Well, some of calls were ignored because of rounding errors: the `time_since_last_profile` was `0.9987283` or the like.

joerick:
Sorry it's taken me so long to get around to this! I've taken a look and it seems the reason the number of calls is increasing is that the first profile event occurs in `profiler.start()`or `profiler.__enter__()`, with a really short duration (e.g. 46µs). I'd rather not have the profiler itself show up in the observations, mostly since it makes these tests confusing.

I'll have a look at some other options, since I agree that there's one too many checks in the code at the moment.

joerick:
I've pushed an update to pyinstrument_cext to avoid the first immediate profile call, which should fix the problem without requiring changes to the tests.

joerick:
Look okay to you?

jd:
Yes, perfect! Thanks @joerick

PR #65, opened 2019-05-18
Fix Windows filename issue for URLs with a query string

obrienmd:
Windows doesn't allow "?" in filenames, which causes the pyinstrument Django middleware to fail for any URL with a query string. 

I've added a simple replace of "?" with "_qs_" to the file path, but I'm not sure it's the right approach; query strings can get pretty long, and the resulting var cuts off at 100 chars (which is reasonable). Another approach might be to write the query string into the HTML output, but that's not such a quick pull request 😄

### DIFF
--- a/pyinstrument/middleware.py
+++ b/pyinstrument/middleware.py
@@ -30,7 +30,7 @@ class ProfilerMiddleware(MiddlewareMixin):
 
             # Limit the length of the file name (255 characters is the max limit on major current OS, but it is rather
             # high and the other parts (see line 36) are to be taken into account; so a hundred will be fine here).
-            path = request.get_full_path().replace('/', '_')[:100]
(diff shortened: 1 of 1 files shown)


joerick:
Oh cool, thanks for this. Any chance we could limit this just to Windows?

obrienmd:
Sure, os.name == 'nt' seems the cleanest way to do this. That being said, Cygwin-compiled Python won't "pass" this test, but that seems less and less used with both native Windows dev improvements and increasing use of WSL (which won't have a problem if they use the Linux file system).

Let me know if that sounds good, or I can try something hackier :)

obrienmd:
Sorry, opened up another pull request... Bad form. https://github.com/joerick/pyinstrument/pull/66

PR #66, opened 2019-05-19
Fix Windows "?" path issue in django middleware

obrienmd:
Detect if running on a Windows platform, and swap ? for _qs_ to avoid Windows issue with ? in filenames.

### DIFF
--- a/pyinstrument/middleware.py
+++ b/pyinstrument/middleware.py
@@ -1,6 +1,7 @@
 from django.http import HttpResponse
 from django.conf import settings
 from pyinstrument import Profiler
+import platform
 import time
 import os
 try:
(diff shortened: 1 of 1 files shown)


pyinstrument/middleware.py
```diff
@@ -32,6 +33,10 @@ def process_response(self, request, response):
             # high and the other parts (see line 36) are to be taken into account; so a hundred will be fine here).
             path = request.get_full_path().replace('/', '_')[:100]
 
+            # Swap ? for _qs_ on Windows, as it does not support ? in filenames.
+            if platform.platform().startswith('Windows'):
```
joerick: Sorry to be difficult (I can see you've been around the houses on this!) but the docs say about `platform.platform()` "The output is intended to be human readable rather than machine parseable"

I think `sys.platform` might be the best approach here. How about `if sys.platform in ['win32', 'cygwin']:`?

obrienmd: Seems totally reasonable to me :) Coming soon!

obrienmd: Updated - thanks for sticking with it!

PR #75, opened 2019-11-26
Add show callback used for determining whether the profiling page should show or not in Django

AlessioBazzanella:
If you want to show the profiling page depending on the request you can define 
`PYINSTRUMENT_SHOW_CALLBACK` as dotted path to a function used for determining 
whether the page should show or not.
You can provide your own function callback(request) which returns True or False 
in your settings.py.

```python
def custom_show_pyinstrument(request):
    return request.user.is_superuser


PYINSTRUMENT_SHOW_CALLBACK = "%s.custom_show_pyinstrument" % __name__
```


joerick:
Nice one @AlessioBazzanella ! This looks good.

PR #88, opened 2020-04-09
Fix #87

### LINKED ISSUE #87: DOC: Add html renderer to the documentation while profiling a chunk of code
In the [Profile a specific chunk of code](https://github.com/santhisenan/pyinstrument/blob/master/README.md#profile-a-specific-chunk-of-code) section of the README, describe HTML renderer. Currently only output_text is described.
```python
print(profiler.output_text(unicode=True, color=True))
```

santhisenan:
Describe HTML renderer in Profile a specific chunk of code section of README


joerick:
Thanks!

Issue #120, opened 2021-02-14
Contributing instructions not working in a vanilla environment


startakovsky:
Hello,

I thought I'd sit down this evening and knock out #119 but I was stopped when the Contributing instructions didn't work for me. After 20 minutes of troubleshooting I figured maybe the community here may have some quick insight to get me up and running. 

I'm running with pyenv, created and activated an environment with Python 3.9.0 and I'm using node v15.5.0.

Here is the output of `python setup.py build` running in my local directory:

```

(pyinstrument) ➜  pyinstrument git:(master)  python setup.py build
running build
running build_py
7.3.0
Running command: npm install

added 1482 packages, removed 7 packages, and audited 1657 packages in 23s

17 vulnerabilities (10 low, 1 moderate, 6 high)

To address all issues, run:
  npm audit fix

Run `npm audit` for details.
Running command: npm run build

> html_js@0.1.0 build
> vue-cli-service build

node:internal/modules/cjs/loader:928
  throw err;
  ^

Error: Cannot find module 'json-parse-even-better-errors'
Require stack:
- /Users/steven.tartakovsky/code/pyinstrument/html_renderer/node_modules/read-pkg/node_modules/parse-json/index.js
- /Users/steven.tartakovsky/code/pyinstrument/html_renderer/node_modules/read-pkg/index.js
- /Users/steven.tartakovsky/code/pyinstrument/html_renderer/node_modules/@vue/cli-service/lib/Service.js
- /Users/steven.tartakovsky/code/pyinstrument/html_renderer/node_modules/@vue/cli-service/bin/vue-cli-service.js
    at Function.Module._resolveFilename (node:internal/modules/cjs/loader:925:15)
    at Function.Module._load (node:internal/modules/cjs/loader:769:27)
    at Module.require (node:internal/modules/cjs/loader:997:19)
    at require (node:internal/modules/cjs/helpers:92:18)
    at Object.<anonymous> (/Users/steven.tartakovsky/code/pyinstrument/html_renderer/node_modules/read-pkg/node_modules/parse-json/index.js:3:18)
    at Module._compile (node:internal/modules/cjs/loader:1108:14)
    at Object.Module._extensions..js (node:internal/modules/cjs/loader:1137:10)
    at Module.load (node:internal/modules/cjs/loader:973:32)
    at Function.Module._load (node:internal/modules/cjs/loader:813:14)
    at Module.require (node:internal/modules/cjs/loader:997:19) {
[...]

PR #121, opened 2021-02-14
Add timeline capabilities to json, html and browser

### LINKED ISSUE #119: timeline in interactive html view
Hi! I love the timeline feature, it really helps me understand the order of operation and get my mind wrapped around execution flow. I get an error when using the `timeline=True` from inside any of the output / open functions except for text output. I guess if this is the place to put a feature request then this is it. I would be happy to submit a PR if it doesn't exist, but I'd want to know what level of effort you would you estimate for this.

startakovsky:
Thanks @joerick for being fast to reply to getting me up and running, and for writing code that makes it straightforward to make/expose this feature to the rest of the renderers all the way to your CLI.

Let me know if things aren't to your liking and we can do a round, otherwise thanks for accepting this.

This closes #119 and #120.

### DIFF
--- a/README.md
+++ b/README.md
@@ -478,7 +478,10 @@ To run pyinstrument from the git repo or a source checkout, you must first run
 
 This compiles the Javascript code needed for the HTML output. You will need
 [node](https://nodejs.org/en/) installed (Node isn't required for the pip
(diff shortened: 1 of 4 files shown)


joerick (approved):
Great! Looks good to me.

PR #124, opened 2021-03-21
Update supported Python versions at top of README

chrish42:
The top of the README used to say that 2.7 and 3.3+ are supported. But if you scroll down to the changelog, it mentions dropping support for 3.5... :-)


joerick:
Ooh, whoops! Thanks :)

Issue #111, opened 2020-12-01
Asyncio applications have a mysterious [SELF] function


AggressivelyMeows:
![](https://nyc3.digitaloceanspaces.com/cerulean/screenshots/2020/12/firefox_Rujk3KwAOv.png)

The first function is labelled as [SELF] and i'm not sure how to interpret it. Is it how long the event loop took to get back to the function or something else?

For context, i'm creating a profiler before and closing it after every request in Sanic. Its a Flask-like asyncio powered web framework.

joerick:
'Self time' is time spent inside the function that's not attributable to any children - so probably something that's happening inside the function directly. I'm not super familiar with async Python, but I _think_ that if your code hit an await, you'd see the flow control return to the run loop in your  sample, which doesn't look like what's happening here. Maybe the time really is spent inside the function `serve`?

AggressivelyMeows:
But then wouldnt it collpase into the serve function time?

AggressivelyMeows:
While attempting to figure this out, I found that PyInstrument only records the first await'ed task during the profile:
![](https://nyc3.digitaloceanspaces.com/cerulean/screenshots/2020/12/firefox_9gYn5DoCA5.png)

There should be the `get_post` function running with the `_` but it just shows the time waster function.

joerick:
Hmm interesting! So the async/await stuff does preserve profiler state to keep call trees clean. Good to know!

wxd:
Does this mean that `pyinstrument` cannot be (easily) combined with async frameworks?

I tried using [`fastapi-profiler`](https://github.com/sunhailin-Leo/fastapi_profiler/blob/main/fastapi_profiler/profiler_middleware/profiler.py#L44), which essentially does:
```python
profiler.start()

try:
    await self.app(scope, receive, wrapped_send)
finally:
    print(self._profiler.output_text(show_all=True))
```

The output seems to contain only framework calls:
```
5.590 run_forever  asyncio/base_events.py:557
`- 5.590 _run_once  asyncio/base_events.py:1784
   |- 5.477 select  selectors.py:553
   |  `- 5.476 kqueue.control  <built-in>:0
   `- 0.112 _run  asyncio/events.py:79
      `- 0.106 coro  starlette/middleware/base.py:36
         `- 0.106 __call__  fastapi_profiler/profiler_middleware/profiler.py:35
            `- 0.106 __call__  starlette/exceptions.py:56
               `- 0.106 __call__  starlette/routing.py:545
[...]

PR #126, opened 2021-04-05
Pyinstrument 4.0 incl. async support

joerick:
I've been thinking a bit more about async since the explorations in #111. I have a few ideas of how to improve things :)

- [x] Scope Profilers to a single async 'context' using [contextvars] (python 3.7+)
   - When execution strays outside of a Profiler's context, that probably means that something in the context did an `await`, and the run loop is busy doing other things before the coroutine can continue. ~Currently this is gathered under a single frame called 'out-of-context'. But perhaps there'd be a way to:~
  - [x] Track 'out-of-context' execution by keeping track of which was the last 'await' statement within the context and attributing the out-of-context execution to that. This would require some changes to the C extension.
  - [x] Add some tests for async use (perhaps using asyncio and trio)
    - [x] Add some assertions on the async time
- [x] Allow multiple profilers to run concurrently on the same thread (assuming that they're scoped to different async contexts)
  - There's a tricky thing to manage regarding the context vars here, as we might get into an issue where there are multiple profilers observing the same execution, say, if an already instrumented app is run from the command line with `pyinstrument`. Ideally, it would be possible for both profilers to observe the same code, but the contextvar thing would mean that the root profiler would see the async context as 'out-of-context'. But I'd argue that the root profiler should see everything. So perhaps the right answer would be to raise an exception if the user attempts to start a profiler in a context that already has one running.
  - [x] raise an exception if the user attempts to start a profiler in a context that already has one running.
- [x] Document the new APIs
  - [x] start a proper documentation site on readthedocs
- [x] Provide a way to configure async support, to allow pyinstrument to profile other async mechanisms, like greenlets. This probably involves disabling the async/context-aware parts of the code.
- [x] Document async support - how it works, and how to configure it

This PR contains the current state of async support, which just understands the difference between async contexts to correctly apportion time in it.

You can try it by doing:

    pip install git+https://github.com/joerick/pyinstrument.git@async

I'd love some feedback on the current version - for example, does it work with async HTTP servers like fast-api or sanic?

<details><summary>Results from my test run</summary>

```python
import asyncio
import pyinstrument
import time


def do_nothing():
    pass


def busy_wait(duration):
    end_time = time.time() + duration

    while time.time() < end_time:
        do_nothing()


async def say(what, when, profile=False):
    if profile:
        p = pyinstrument.Profiler()
        p.start()

    busy_wait(0.1)
    sleep_start = time.time()
    await asyncio.sleep(when)
    print(f'slept for {time.time() - sleep_start:.3f} seconds')
    busy_wait(0.1)

    print(what)
    if profile:
        p.stop()
        p.print(show_all=True)

loop = asyncio.get_event_loop()

loop.create_task(say('first hello', 2, profile=True))
loop.create_task(say('second hello', 1))
loop.create_task(say('third hello', 3))

loop.run_forever()
loop.close()
```

```text
slept for 1.003 seconds
second hello
slept for 2.005 seconds
first hello

  _     ._   __/__   _ _  _  _ _/_   Recorded: 18:27:13  Samples:  503
 /_//_/// /_\ / //_// / //_'/ //     Duration: 2.206     CPU time: 0.502
/   _/                      v3.4.1

Program: examples/async_experimentation.py

2.205 Async context  None:0
├─ 2.005 <out-of-context>  <out-of-context>:0
└─ 0.200 say  async_experimentation.py:17
   └─ 0.199 busy_wait  async_experimentation.py:10
      ├─ 0.115 [self]  
      └─ 0.065 time  <built-in>:0


slept for 3.005 seconds
third hello
```

Notice that the durations of busy_wait and out-of-context (here, asyncio.sleep) now match the code! Before this PR, you can see how this code performs before this PR in this post: https://github.com/joerick/pyinstrument/issues/111#issuecomment-813021047 

</details>


joerick:
Alrighty... I think this is ready for prime time!

PR #127, opened 2021-04-09
docs: update code example with interval

miketheman:
Add details on how to use the `interval` input to Profiler()

Thanks for this library! I was a little confused as to why I didn't see any samples when running with the "inline code" option, so I figured this might help someone else.

I ran the same code via CLI vs inline code profiling, and the CLI produced results, so I'm not sure if this is the right answer/solution, but figured I'd surface it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,6 +105,18 @@ print(profiler.output_text(unicode=True, color=True))
 (You can omit the `unicode` and `color` flags if your output/terminal does
 not support them.)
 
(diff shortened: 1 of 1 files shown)


joerick:
Thanks. Yes, this is correct. To be honest, the library could use a lot more documentation, but I wasn't sure how many people would be using the Python interface when I originally wrote it. Turns out, quite a few! So I need to document it a bit better. But I want to crack async profiling first :)

Thanks for this.

PR #135, opened 2021-06-28
Bugfix: pass processor_options down to JSONRenderer

danvk:
I was running with `--load-prev` with `-r html` and trying to figure out why `--hide-regex` wasn't doing anything. This is why.


joerick:
Ahhhhh, good catch. Thank you!

PR #141, opened 2021-08-08
Optimize callback routines when the API is available

jnsquire:
I noticed some deprecation warnings when building pyinstrument again cpython HEAD, and decided to go all the way and use the new optimized function invocation options available in 3.9, when possible. No idea how much this affects the actual performance.

### DIFF
--- a/pyinstrument/low_level/stat_profile.c
+++ b/pyinstrument/low_level/stat_profile.c
@@ -273,22 +273,12 @@ call_target(ProfilerState *pState, PyFrameObject *frame, int what, PyObject *arg
 static PyObject *
 call_target(ProfilerState *pState, PyFrameObject *frame, int what, PyObject *arg)
 {
-    PyObject *whatstr;
-    PyObject *result;
-
-    whatstr = whatstrings[what];
-    Py_INCREF(whatstr);
-
     PyFrame_FastToLocals(frame);
-    Py_INCREF(frame);
-
-    PyObject *callargs[4] = { NULL, (PyObject *) frame, whatstr, arg == NULL ? Py_None : arg };
-
-    result = PyObject_Vectorcall(pState->target, callargs + 1, 3 | PY_VECTORCALL_ARGUMENTS_OFFSET, NULL);
+    PyObject *callargs[4] = { NULL, (PyObject *) frame, whatstrings[what], arg == NULL ? Py_None : arg };
+    PyObject *result = PyObject_Vectorcall(pState->target, callargs + 1, 3 | PY_VECTORCALL_ARGUMENTS_OFFSET, NULL);
     PyFrame_LocalsToFast(frame, 1);
     if (result == NULL)
         PyTraceBack_Here(frame);
-
 
     return result;
 }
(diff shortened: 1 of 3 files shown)


pyinstrument/low_level/stat_profile.c
```diff
@@ -265,6 +269,30 @@ call_target(ProfilerState *pState, PyFrameObject *frame, int what, PyObject *arg
...
+
+    whatstr = whatstrings[what];
+    Py_INCREF(whatstr);
+
+    PyFrame_FastToLocals(frame);
+    Py_INCREF(frame);
+
+    PyObject *callargs[4] = { NULL, (PyObject *) frame, whatstr, arg == NULL ? Py_None : arg };
+
+    result = PyObject_Vectorcall(pState->target, callargs + 1, 3 | PY_VECTORCALL_ARGUMENTS_OFFSET, NULL);
```
joerick, line 287: I wonder if [PyObject_CallFunctionObjArgs](https://docs.python.org/3/c-api/call.html#c.PyObject_CallFunctionObjArgs) would be neater here?

pyinstrument/low_level/stat_profile.c
```diff
@@ -265,6 +269,30 @@ call_target(ProfilerState *pState, PyFrameObject *frame, int what, PyObject *arg
...
+call_target(ProfilerState *pState, PyFrameObject *frame, int what, PyObject *arg)
+{
+    PyObject *whatstr;
+    PyObject *result;
+
+    whatstr = whatstrings[what];
+    Py_INCREF(whatstr);
+
+    PyFrame_FastToLocals(frame);
+    Py_INCREF(frame);
```
joerick, line 283: I don't think these increfs are needed any more, I believe they were there because the PyTuple_SET_ITEM steals a refcount.

joerick (commented):
Interesting, thanks @jnsquire. I can't find the deprecation warnings you mention, could you find/link/paste them? I'm interested in which APIs are deprecated.

I'm not sure if this method will be much quicker, because this C function always calls into Python code, but maybe it skips the malloc of the args tuple? That would be nice.

pyinstrument/low_level/stat_profile.c
jnsquire: There's a blurb about the deprecation of the PyEval_CallObject* routines in favor of PyObject_CallObject* in the Python 3.9 release notes. The vectorcall variant is supposed to be the fastest when you can arrange the preconditions right, since it doesn't require construction of a tuple. The extra NULL slot at the beginning of the callargs array allows reused of the structure when using bound methods, I think.

pyinstrument/low_level/stat_profile.c
jnsquire: Yeah, I wondered about that. I'll remove those.

pyinstrument/low_level/stat_profile.c
joerick: Ah, cool, I'll take a look. Btw PyObject_CallFunctionObjArgs is one of the new APIs, I think

jnsquire:
I saw a couple of test failures in the testcases that use greenlet with this change in place (incorrect number of records with sleep both times), but that went away after I rebuilt my venv. Not sure if that was a fluke or not.

pyinstrument/low_level/stat_profile.c
joerick: Oh, I see, the vectorcall calling convention allows for more efficient bound method calling. Cool, let's keep this.

joerick:
No I was checking it worked with Python 3.10, which was merged recently. I just have a couple of experiments to run locally before merging...

joerick:
Hi @jnsquire, I just added a few tweaks to this, we can drop the PyEval calls entirely because pyinstrument is 3.7+. Let me know if this looks okay to you?

PR #147, opened 2021-09-05
Use PyFrame_GetCode when it's available

jnsquire:
Build fails on latest CPython version due to frame->f_code deprecation. Use new API to fetch the code object when it's present. Also fix the Python version detection macros used, after I noticed they were wrong.


joerick:
Thanks @jnsquire. I spotted a memory leak (no code decref if `code->co_flags & 0x80` is false) so I pushed a fix to that. Let me know if it still makes sense to you.

jnsquire:
Yeah, that seem's fine to me.

PR #151, opened 2021-09-27
Fix typos discovered by codespell

cclauss:
[`codespell --ignore-words-list=vas --skip="*.js"`](https://pypi.org/project/codespell)


joerick:
thanks!

Issue #118, opened 2021-02-09
Would it be possible to show the class name in the results ?


altimore:
Hello, 

First many thanks for the work you put in this tool, it was the one i was dreaming of.

I'm using it to find bottlenecks in Django applications.

In this case i got multiple classes having a method called get_total(), would it be possible to display the class name to differentiate the different get_total() functions like Invoice.get_total(), InvoiceLine.get_total() or InvoiceQuerySet.get_total() in the results instead of just get_total.
It would be more convenient than having to check which is which in the source file during the reading.

Regards, 

Alti.

joerick:
Yes, this would be really nice. It's not something that is naturally available via the [native Python `frame` interface](https://docs.python.org/3/library/inspect.html) but I think it's possible to 'guess' by looking at the type of an argument called `self` in the scope of the function. We'd have to be a little careful not to make it too slow, but I think there should be a way.

If you or somebody else is interested in working on a PR, the place to try adding it would be https://github.com/joerick/pyinstrument/blob/master/pyinstrument/profiler.py#L93, adding it to the first token in the identifier :)

joerick:
This now works in v4.4.0!

PR #154, opened 2021-10-06
Use qualified name for functions

naspeh:
So class methods will have `<class name>.<method name>`, etc.

Related to #118 

----

@joerick, thank you for the great profiler!

### DIFF
--- a/pyinstrument/stack_sampler.py
+++ b/pyinstrument/stack_sampler.py
@@ -1,5 +1,8 @@
 from __future__ import annotations
 
+
+import functools
+import gc
 import threading
 import timeit
 import types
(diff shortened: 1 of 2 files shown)


pyinstrument/stack_sampler.py
```diff
@@ -160,6 +163,17 @@ def get_stack_sampler() -> StackSampler:
     return thread_locals.stack_sampler
 
 
+@functools.cache
+def get_qualified_name(code):
+    name = code.co_name
+    refs = gc.get_referrers(code)
```
naspeh, line 169: Inspired by https://stackoverflow.com/a/52762678

PR #156, opened 2021-10-09
Improve frame identifier if self or cls exist

naspeh:
Another solution for #118

I decided to create a separate PR with the solution discussed in #154


joerick:
I wasn't able to push to this PR with the C extension work, so I've opened a new one at #166.

joerick:
Work continues in #166

Issue #153, opened 2021-10-05
Jupyter integration


mlucool:
Hi,

Would you be willing to accept a PR that adds Jupyter magic integration? It would look something like this: 
![image](https://user-images.githubusercontent.com/1813603/136089667-bda54813-8a29-406c-8d3a-b1d973d44031.png)

Doing this would add IPython as a dependency (or optional dependency).

Thanks in advance for considering this!

PR #157, opened 2021-10-13
Add support for IPython/Jupyter notebook magics

mlucool:
Example usage (works in ipython/notebook/lab):
```
%load_ext pyinstrument
```

```
%%pyinstrument
def a():
    b()
    c()
def b():
    d()
def c():
    d()
def d():
    e()
def e():
    time.sleep(1)
a()
```
image

Extra care is taken to not profile jupyter running the code, which
is the majority of the complexity of this commit.

Fixes: #153

### DIFF
--- a/test/test_overflow.py
+++ b/test/test_overflow.py
@@ -1,3 +1,4 @@
+import faulthandler
 import inspect
 import sys
 import time
@@ -50,7 +51,6 @@ def test_console(deep_profiler_session):
     ConsoleRenderer().render(deep_profiler_session)
 
 
-# html now uses the json renderer, so it's xfail too.
 def test_html(deep_profiler_session):
     HTMLRenderer().render(deep_profiler_session)
 
--- a/test/test_overflow.py
+++ b/test/test_overflow.py
@@ -1,4 +1,3 @@
-import faulthandler
 import inspect
 import sys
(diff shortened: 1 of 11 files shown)


joerick:
I'd like to have a go at this, seems like there might be a way to incorporate a test based on your starting point.

joerick:
So. I have the IPython test working, but for some reason, even importing the IPython codebase on Windows causes a memory/stack overflow crash in the overflow test. No clue what's going on there, but it's probably not worth the time to debug. IPython is more of an application than a library, I expect there's some global/system manipulation going on that's causing an issue. I'll just sidestep the whole thing and run the IPython tests in their own process.

test/test_overflow.py
```diff
@@ -1,3 +1,4 @@
+import faulthandler
```
joerick, line 1:
```suggestion
```

joerick:
Cool, I think this is ready. Does it look good to you @mlucool?

mlucool:
This looks great! Thanks for all the work on this!

mlucool:
Thanks for merging @joerick. When do you expect to release this?

joerick:
Should be in the next few days, I want to get #160 in and then we should be good to go

PR #158, opened 2021-10-16
Add Option to configure renderers in Django settings

kracekumar:
### Why?

- HTMLRenderer is a default renderer for Django App. 
- It'll be useful to support exporting the profile data for various formats to generate alternate visualization graphs like flame graph and use with other applications like speedscope.

### Changes

- Supports a new Django settings variable `PYINSTRUMENT_RENDERER` which takes in the path of the renderer to use.
- The middleware produces the output in the configured path. The file extension can only be `json, HTML, txt` for now. I'm happy to make the extension configurable too.

Let me know If I need to add test cases and modify the docs. 

Thanks.

### DIFF
--- a/pyinstrument/middleware.py
+++ b/pyinstrument/middleware.py
@@ -25,7 +25,6 @@ def get_renderer_and_extension(path):
         except ImportError as exc:
             print("Unable to import the class: %s" % path)
             raise exc
-            sys.exit(-1)
 
     else:
         renderer = HTMLRenderer()
--- a/pyinstrument/middleware.py
+++ b/pyinstrument/middleware.py
@@ -8,8 +8,8 @@ from django.http import HttpResponse
 from django.utils.module_loading import import_string
 
 from pyinstrument import Profiler
-from pyinstrument.renderers.html import HTMLRenderer
 from pyinstrument.renderers import JSONRenderer
+from pyinstrument.renderers.html import HTMLRenderer
 
 try:
     from django.utils.deprecation import MiddlewareMixin
@@ -18,8 +18,7 @@ except ImportError:
 
 
 def get_renderer_and_extension(path):
-    """Return the renderer instance and output file extension.
-    """
+    """Return the renderer instance and output file extension."""
     if path:
         renderer = import_string(path)()
     else:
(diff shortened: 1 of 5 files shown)


pyinstrument/renderers/base.py
```diff
@@ -29,6 +29,10 @@ class Renderer:
     Dictionary containing processor options, passed to each processor.
     """
 
+    output_file_extension: str = "txt"
```
kracekumar, line 32: I have added the `output_file_extension` attribute.

pyinstrument/middleware.py
```diff
@@ -41,8 +61,10 @@ def process_response(self, request, response):
         if hasattr(request, "profiler"):
             profile_session = request.profiler.stop()
 
-            renderer = HTMLRenderer()
-            output_html = renderer.render(profile_session)
+            configured_renderer = getattr(settings, "PYINSTRUMENT_PROFILE_DIR_RENDERER", None)
```
kracekumar, line 64: Keeping this here since  this renderer is used in the `if/else` block to produce output.

kracekumar (commented):
I have used `print` statement to print error message. Let me know if I need to use logger.

pyinstrument/middleware.py
```diff
@@ -16,6 +17,25 @@
...
 
+def get_renderer_and_extension(path):
+    """Return the renderer instance and output file extension."""
+    if path:
+        try:
+            renderer = import_string(path)()
+        except ImportError as exc:
+            print("Unable to import the class: %s" % path)
+            raise exc
+            sys.exit(-1)
```
joerick, line 28:
```suggestion
```

joerick (approved):

pyinstrument/middleware.py
```diff
@@ -16,6 +17,25 @@
...
+            raise exc
+
+    else:
+        renderer = HTMLRenderer()
+
+    if isinstance(renderer, Renderer):
+        return renderer, renderer.output_file_extension
+    else:
+        print("Renderer should subclass: %s. Using HTMLRenderer" % Renderer)
+        return HTMLRenderer(), HTMLRenderer.output_file_extension
```
kracekumar, line 36: Pyright was pointing out this can return None. So when the renderer is not a subclass of Renderer, we will be using HTMLRenderer. Let me know if this is fine?

joerick: I think it would be better to raise an error, instead. I'll update it.

joerick: I've made the update. For some reason I wasn't able to push to this branch. I've pushed it to https://github.com/joerick/pyinstrument/commit/69b195b6649344b04861df4ca54f8b372a5c88e0 - if you could cherry-pick that here, I think we're good to go.

joerick:
This was merged in #163

kracekumar:
Thanks a lot! Couldn't do complete earlier.

PR #160, opened 2021-10-25
Add speedscope renderer

### LINKED ISSUE #89: Feature Request: Add support for speedscope
Dear Authors,
Thank you so much for this package! It's very useful. I was hoping you could add support for speedscope like py-spy does, which is a very detailed flame chart application (hosted on https://www.speedscope.app/)

goxberry:
This pull request adds a renderer that outputs JSON conforming to  [speedscope](https://github.com/jlfwong/speedscope)'s JSON schema; in particular, speedscope's "evented" format is used.

This implementation is very rough, and intended as a request for comment. I expect to revise this pull request based on feedback from the project maintainer(s) to better conform with project style.

Closes #89.

### DIFF
--- a/test/test_profiler.py
+++ b/test/test_profiler.py
@@ -182,138 +182,86 @@ def test_speedscope_output():
     speedscope_frame_fields = tuple([field.name for field in dataclasses.fields(SpeedscopeFrame)])
     for (function_name, frame_index) in distinct_functions_called.items():
         for frame_field in speedscope_frame_fields:
-            assert frame_field in speedscope_frame_list[frame_index], (
-                f"Field named '{frame_field}' not in Speedscope output"
-                f"at 'shared.frames[{frame_index}]'"
-            )
-        frame_index_name = speedscope_frame_list[frame_index]["name"]
-        assert frame_index_name == function_name, (
-            f"Speedscope output 'shared.frames[{frame_index}].name' has value "
-            f"{frame_index_name} != {function_name}"
-        )
+            assert frame_field in speedscope_frame_list[frame_index]
+        assert speedscope_frame_list[frame_index]["name"] == function_name
 
     speedscope_profile_list = output["profiles"]
     assert len(speedscope_profile_list) == 1
     speedscope_profile = speedscope_profile_list[0]
     speedscope_profile_fields = ("type", "name", "unit", "startValue", "endValue", "events")
     for profile_field in speedscope_profile_fields:
-        assert (
-            profile_field in speedscope_profile
-        ), f"Field named '{profile_field}' not in Speedscope output at 'profiles[0]'"
+        assert profile_field in speedscope_profile
 
     # speedscope_profile["endValue"] is not tested because a fake_time mock
     # timer is used to replace time.time, and this mock causes session.duration
     # to differ from self._event_time just before exiting SpeedscopeRenderer.render
-    start_time_in_seconds = 0.0
-    event_time_in_seconds = start_time_in_seconds
     assert speedscope_profile["type"] == "evented"
     assert speedscope_profile["unit"] == "seconds"
-    assert speedscope_profile["startValue"] == start_time_in_seconds
-
-    """Two helper functions for constructing a timeline and testing it"""
-
-    def update_event_time_on_sleep_close(time_increment_in_seconds: float):
-        """Updates running total of event time in seconds and returns updated value"""
-        nonlocal event_time_in_seconds
-        event_time_in_seconds += time_increment_in_seconds
-        return event_time_in_seconds
-
-    def get_approx_abs_tol(time_in_seconds: float):
-        """Computes absolute tolerance for given time value to account for CI time jitter"""
-        uncertainty_per_quarter_second = 0.1
-        absolute_tolerance = (time_in_seconds / 0.25) * uncertainty_per_quarter_second
-        return absolute_tolerance
-
-    speedscope_event_list = speedscope_profile["events"]
-    sleep_a_time_in_seconds = 0.25
-    sleep_b_time_in_seconds = 0.5
+    assert speedscope_profile["startValue"] == 0.0
 
     output_event_tuple = (
         SpeedscopeEvent(
             SpeedscopeEventType.OPEN,
-            event_time_in_seconds,
+            0.0,
             distinct_functions_called["test_speedscope_output"],
         ),
         SpeedscopeEvent(
             SpeedscopeEventType.OPEN,
-            event_time_in_seconds,
+            0.0,
             distinct_functions_called["long_function_a"],
         ),
         SpeedscopeEvent(
             SpeedscopeEventType.OPEN,
-            event_time_in_seconds,
+            0.0,
             distinct_functions_called["sleep"],
         ),
         SpeedscopeEvent(
             SpeedscopeEventType.CLOSE,
-            update_event_time_on_sleep_close(sleep_a_time_in_seconds),
+            0.25,
             distinct_functions_called["sleep"],
         ),
         SpeedscopeEvent(
             SpeedscopeEventType.CLOSE,
-            event_time_in_seconds,
+            0.25,
             distinct_functions_called["long_function_a"],
         ),
         SpeedscopeEvent(
             SpeedscopeEventType.OPEN,
-            event_time_in_seconds,
+            0.25,
             distinct_functions_called["long_function_b"],
         ),
         SpeedscopeEvent(
             SpeedscopeEventType.OPEN,
-            event_time_in_seconds,
+            0.25,
             distinct_functions_called["sleep"],
         ),
         SpeedscopeEvent(
             SpeedscopeEventType.CLOSE,
-            update_event_time_on_sleep_close(sleep_b_time_in_seconds),
+            0.75,
             distinct_functions_called["sleep"],
         ),
         SpeedscopeEvent(
             SpeedscopeEventType.CLOSE,
-            event_time_in_seconds,
+            0.75,
             distinct_functions_called["long_function_b"],
         ),
         SpeedscopeEvent(
             SpeedscopeEventType.CLOSE,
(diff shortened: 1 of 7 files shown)


joerick (commented):
Brilliant, this is looking really good, @goxberry! I appreciate the effort you've put into this :)

Would you be able to add the Speedscope renderer to the overflow test and write a simple test for the speedscope output?

I'm thinking something like:

- test starts profiler
- test calls function a
  - function a calls function b
    - function b sleeps (you can use 'fake time' for this, see other tests)
    - function b returns
  - function a returns
- test stops profiler
- test renders speedscope JSON
- test loads speedscope JSON
- test checks a few things about the JSON, I suppose, a few things about events and metadata. I'm not sure of the specifics, perhaps the events should be [open, open, close, close]?

goxberry:
@joerick Thanks! And thank you for your review and for writing this profiler! Your design has made it straightforward to add this new feature. As you've suggested, I've added `SpeedscopeRenderer` to the tests in `test/test_overflow.py` and `test/test_profiler.py`. Because the functions `long_function_a` and `long_function_b` within `test/test_profiler.py` call the `sleep` function, calling both `long_function_a` and `long_function_b` suffices to generate an event timeline with nested function calls, so I elected to make the `test_speedscope_output` test similar to the other `test_*_output` tests for simplicity.

As a result of testing, I noticed something peculiar in `SpeedscopeRenderer` and wanted to get your insight about what might be happening. When I run `test_profiler.test_speedscope_output` in a debugger and:

* set a breakpoint at `test/test_profiler.py:149` and continue until I reach it
* step into `pyinstrument/renderers/speedscope.py`
* set another breakpoint at `pyinstrument/renderers/speedscopy.py:220` and continue until I reach it

I notice that the value of `session.duration` at that line is less than the value of `self._event_time`, and I'm not sure why `session.duration < self._event_time`; I would have expected `session.duration >= self._event_time`.

You should be able to reproduce this phenomenon yourself by running in a terminal:
```bash
> pytest --trace ${PATH_TO_PYINSTRUMENT_REPO}/test/test_profiler.py -k "test_speedscope_output"
(Pdb) b 149
(Pdb) c
(Pdb) s    # should step into speedscope.py.__init__() here
(Pdb) b 220
(Pdb) c
(Pdb) p session.duration     # on my machine, this command prints the value 0.00041103363037109375
(Pdb) p self._event_time      # on my machine, this command prints 0.75, as expected, within a tolerance of 0.3
```

where the first `>` denotes a shell prompt, and the `(Pdb)` denotes a `pdb` debugger prompt.

Do you have any insight as to why this behavior might occur?

joerick:
Hi @goxberry! Yes, I can offer an explanation. The session.duration is computed by calling `time.time()` before and after profiling. But we are using 'fake time', which is defined here:

https://github.com/joerick/pyinstrument/blob/96dcb80ea480a93fabe9912a9d0089da3485f088/test/fake_time_util.py#L23-L32

This monkeypatches `time.sleep` and some internal machinery, but not `time.time`. So in the reality the session duration is correct, very little time has passed, but the test has a simulated reality, where time has passed!

If this is causing a problem, you could edit the `fake_time` function to also monkeypatch `time.time` (returning `fake_clock.get_time` should do the trick). Or we can just ignore the difference :)

goxberry:
Thanks for the explanation! 

> If this is causing a problem, you could edit the fake_time function to also monkeypatch time.time (returning fake_clock.get_time should do the trick). Or we can just ignore the difference :)

In the interest of pragmatism, let's ignore the difference. I'll revert the change in `SpeedscopeRenderer` regarding `session.duration` vs `self._event_time` and remove the assertion checking the `endValue` field in the Speedscope output.

test/test_profiler.py
```diff
@@ -138,6 +140,182 @@ def test_json_output():
...
+        "sleep": 2,
+        "long_function_b": 3,
+    }
+    assert len(speedscope_frame_list) == len(distinct_functions_called)
+    speedscope_frame_fields = tuple([field.name for field in dataclasses.fields(SpeedscopeFrame)])
+    for (function_name, frame_index) in distinct_functions_called.items():
+        for frame_field in speedscope_frame_fields:
+            assert frame_field in speedscope_frame_list[frame_index], (
+                f"Field named '{frame_field}' not in Speedscope output"
+                f"at 'shared.frames[{frame_index}]'"
```
joerick, line 187: I really appreciate the craft and effort in these assert messages! But when you're running a test with pytest, and an assert fails, pytest will output a lot of useful information that is relevant to that fail, including the contents of variables etc. So it's probably not worth including that information as an assert message, I think. Also, it will help the test be easier to read.

joerick: Removing them will also reduce the need for a lot of local variables, which should help readability a bit.

test/test_profiler.py
```diff
@@ -138,6 +140,182 @@ def test_json_output():
...
+        """Updates running total of event time in seconds and returns updated value"""
+        nonlocal event_time_in_seconds
+        event_time_in_seconds += time_increment_in_seconds
+        return event_time_in_seconds
+
+    def get_approx_abs_tol(time_in_seconds: float):
+        """Computes absolute tolerance for given time value to account for CI time jitter"""
+        uncertainty_per_quarter_second = 0.1
+        absolute_tolerance = (time_in_seconds / 0.25) * uncertainty_per_quarter_second
+        return absolute_tolerance
```
joerick, line 225: We can use pytest.approx(rel=...) for this, instead of the function call. It would be `uncertainty_per_second = 0.4`... which seems very large! Are you sure we need this? I'd have thought that the fake_time would make CI jitter irrelevant.

test/test_profiler.py
```diff
@@ -138,6 +140,182 @@ def test_json_output():
...
+            distinct_functions_called["long_function_a"],
+        ),
+        SpeedscopeEvent(
+            SpeedscopeEventType.OPEN,
+            event_time_in_seconds,
+            distinct_functions_called["sleep"],
+        ),
+        SpeedscopeEvent(
+            SpeedscopeEventType.CLOSE,
+            update_event_time_on_sleep_close(sleep_a_time_in_seconds),
```
joerick, line 249: Personally, I'd prefer to keep tests pretty explicit. I think it's okay to hardcode things in tests, rather than computing stuff like this.

joerick: So here I'd just put literal `0`, or `sleep_a_time_in_seconds`, or `sleep_a_time_in_seconds + sleep_b_time_in_seconds`

test/test_profiler.py
```diff
@@ -138,6 +140,182 @@ def test_json_output():
...
+                f"Speedscope output 'profiles[0].events[{event_index}]' lacks "
+                f"a field named '{event_field}'"
+            )
+
+        output_event_type = output_event_tuple[event_index].type.value
+        speedscope_event_type = speedscope_event["type"]
+        assert speedscope_event["type"] == output_event_tuple[event_index].type.value, (
+            f"Speedscope output 'profiles[0].events[{event_index}].type has "
+            f"value {speedscope_event_type} != {output_event_type}"
+        )
```
joerick, line 298: I think this section will be quite a lot simplified by removing the assert statements (and associated local variables), as mentioned above.

test/test_profiler.py
joerick: Maybe try doing pytest.approx() without setting abs or rel, I think it might work...

joerick (commented):
Hi @goxberry! Thanks for adding the test. I have a couple requests :)

test/test_profiler.py
goxberry: Works for me.

test/test_profiler.py
goxberry: Those comparisons don't require tolerances on my machine. I pattern-matched the tolerances to the following section of code, which motivated the 0.1s absolute tolerance per 0.25s of fake time:
https://github.com/joerick/pyinstrument/blob/c2e98aaca9581dd9a25373c51b704bf45fce62e8/test/test_profiler.py#L105-L108.

I'll remove the tolerances. All of the floating point numbers involved should be exactly representable per the IEEE-754 standard, so as long as the fake times aren't affected by external factors, tolerances should be unnecessary.

test/test_profiler.py
goxberry: I replaced those calls with literals and removed the local variables re: sleep time.

test/test_profiler.py
goxberry: Removed assertion messages and associated local variables in [`97925fb2`](https://github.com/joerick/pyinstrument/pull/160/commits/97925fb284a34a111b7b99bbef8cd0bb72e7c584).

test/test_profiler.py
goxberry: Replaced function calls with literals in [`97925fb2`](https://github.com/joerick/pyinstrument/pull/160/commits/97925fb284a34a111b7b99bbef8cd0bb72e7c584).

test/test_profiler.py
goxberry: Tolerances and local methods removed in [`97925fb2`](https://github.com/joerick/pyinstrument/pull/160/commits/97925fb284a34a111b7b99bbef8cd0bb72e7c584).

test/test_profiler.py
goxberry: Removed assertion messages and associated local variables in [`97925fb2`](https://github.com/joerick/pyinstrument/pull/160/commits/97925fb284a34a111b7b99bbef8cd0bb72e7c584).

goxberry:
Rebased on `main`.

test/test_profiler.py
joerick: Ah, gotcha, thanks for the explanation! Those tolerances look like a mistake, a hangover from the time before I had fake_time implemented. Sorry they misled you!

RyannDaGreat:
Whoah :O Thank you guys so much!! This is so cool!! :D

PR #161, opened 2021-10-28
Add developer tools used in CI checks to requirements and add documentation on their use in this project

goxberry:
As part of work on #160, I realized I wasn't running certain checks the CI was doing, and these checks caused CI testing failures. This pull request attempts to head off that situation in the future by:

* adding documentation on how to use `pre-commit`, `isort`, `black`, and a Python `pyright` wrapper to diagnose and fix style errors and static analysis bugs
* adds these tools to the project developer requirements file
* also adds numpy to the developer requirements file because it's used in an example

### DIFF
--- a/README.md
+++ b/README.md
@@ -343,7 +343,9 @@ To setup a dev environment:
 
     virtualenv --python=python3 env
     . env/bin/activate
+    pip install --upgrade pip
     pip install -r requirements-dev.txt
+    pre-commit install --install-hooks
 
 To get some sample output:
 
@@ -352,6 +354,10 @@ To get some sample output:
 To run the tests:
 
     pytest
+
+To run CI checks locally:
+
+    pre-commit run --all-files
 
 ### The HTML renderer Vue.js app
 
--- a/README.md
+++ b/README.md
@@ -355,9 +355,24 @@ To run the tests:
 
     pytest
 
-To run CI checks locally:
+To run linting checks locally:
 
     pre-commit run --all-files
+
+Some of the pre-commit checks, like `isort` or `black`, will auto-fix
+the problems they find. So if the above command returns an error, try
+running it again, it might succeed the second time :)
+
+Running all the checks can be slow, so you can also run checks
+individually, e.g., to format source code that fails `isort` or `black`
+checks:
+
+    pre-commit run --all-files isort
+    pre-commit run --all-files black
+
+To diagnose why `pyright` checks are failing:
+
+    pre-commit run --all-files pyright
 
 ### The HTML renderer Vue.js app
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -353,6 +355,19 @@ To run the tests:
...
+
+To format source code that fails `isort` or `black` checks:
+
+    isort --skip env . --skip pyinstrument/vendor .
+    black .
+
+To diagnose why `pyright` checks are failing:
+
+    PYRIGHT_PYTHON_FORCE_VERSION=1.1.134 pyright
+
```
joerick, line 370:
```suggestion
To run linting checks locally:

    pre-commit run --all-files

Some of the pre-commit checks, like `isort` or `black`, will auto-fix the problems they find. So if the above command returns an error, try running it again, it might succeed the second time :)

Running all the checks can be slow, you can also run checks individually, e.g. to format source code that fails `isort` or `black` checks:

    pre-commit run --all-files isort
    pre-commit run --all-files black

To diagnose why `pyright` checks are failing:

    pre-commit run --all-files pyright
```

requirements-dev.txt
```diff
@@ -10,3 +10,12 @@ sphinxcontrib-programoutput==0.17
...
 nox
+numpy # used by an example
+pre-commit # used to run pre-commit hooks
+# ----------------------------------------------------------
+# packages useful for diagnosing & fixing CI check failures
+#------------------------------------------------------------
+isort==5.8.0
+black==20.8b1
+codespell==2.1.0
+pyright # for local debugging only; not for production use in CI
```
joerick, line 21: It probably makes sense just to use these through pre-commit, especially pyright, which pre-commit can install from the official npm package rather than this 3rd party one.

```suggestion
```

joerick (commented):
Thanks @goxberry ! I was looking for this information earlier in fact!

I think it's perhaps easier to use these tools through the pre-commit tool, which handles the installation and configuration. So I've suggested changing it to that.

requirements-dev.txt
goxberry: Sounds good to me; removed in [`235795ad`](https://github.com/joerick/pyinstrument/pull/161/commits/235795ad4551635819ad04719b19b6160dcc050f).

README.md
goxberry: Made this change in [`235795ad`](https://github.com/joerick/pyinstrument/pull/161/commits/235795ad4551635819ad04719b19b6160dcc050f); the only additional edit I made to the text above was changing the substring "slow, you" to "slow, so you".

goxberry:
Rebased on more recent version of `main`.

joerick (approved):

joerick:
Thanks @goxberry !

PR #166, opened 2021-11-14
Include class name in function name for methods and class methods

joerick:
Supersedes #156.


joerick:
Soooo I have this working, and I'm 99% happy with it. However, there's still a problem in async support that's causing me issues. This is a profile of an async test.

```
  _     ._   __/__   _ _  _  _ _/_   Recorded: 10:30:22  Samples:  8
 /_//_/// /_\ / //_// / //_'/ //     Duration: 0.001     CPU time: 0.001
/   _/                      v4.0.4

Program: /opt/hostedtoolcache/Python/3.8.12/x64/bin/pytest

0.600 _UnixSelectorEventLoop._run_once  asyncio/base_events.py:1784
├─ 0.500 NoneType._run  asyncio/events.py:79
│  └─ 0.500 async_wait  test/test_profiler_async.py:92
│     └─ 0.500 sleep  asyncio/tasks.py:641
│        └─ 0.500 [await]  
└─ 0.100 Handle._run  asyncio/events.py:79
   └─ 0.100 async_wait  test/test_profiler_async.py:92
      └─ 0.100 FakeClockAsyncio.sleep  test/fake_time_util.py:46
```

async_wait should be one frame in the stack, it's doing some sync work followed by some async work. But it's being recorded as two frames, one under Handle._run, the other under NoneType._run.

The reason for this is in the asyncio code base:

https://github.com/python/cpython/blob/b7360ae395e9e633d384d16064c5dc04a9841e19/Lib/asyncio/events.py#L78-L95

We're checking the type of `self`, but this code sets `self` to None! Normally I'd say that this wasn't an issue, but it really is an issue because this is in the asyncio library so breaks pyinstrument's async support.

Not really sure what to do to work around this, going to have to think some more on it...

joerick:
Had an idea to work around this. The class name is not required for the 'identifier' purpose of the frame identifer, it's more 'execution information'. So we could store this information as a separate field in the frame identifier.... it's no longer a frame identifier, it's more 'frame info'. But Frame objects can still be created from this string and Renderers can use it how they like.

Gonna be a decent refactor! But could be useful for future features too.

joerick:
Superceded by #203

PR #172, opened 2021-12-17
Update guide.md to prevent error.

koaning:
If you copy/paste the original codeblock you get an error because `time` isn't imported. 

Figured I'd make a small PR to improve it. 

Great library by the way!


joerick:
Nice one, thanks @koaning !

Issue #171, opened 2021-12-16
Replace times with percentage?


willmcgugan:
Thanks for this project. I've found it really easy to use and I love the output.

Would it be possible to show the times as a percentage of the total runtime rather than in seconds? I would find that a more useful metric generally. Perhaps it could be enabled with a command line switch?

joerick:
Hi @willmcgugan ! Thanks for the kind words :)

I guess you're talking about the Console renderer? Yes I think we could add at to the output, under a renderer option / command line switch. Perhaps it's time to generalise the renderer options interface on the command line, too. I'll have a think about this.

joerick:
Now possible by doing `pyinstrument -p time=percent_of_total`!

Issue #132, opened 2021-06-09
Include percentages in the report


asmeurer:
It would be useful if the report showed the percentage of the time spent relative to the other frames with the same parent in addition to the absolute time. 

I looked at the code, and the frame objects do not seem to reference their parents, only their children, so this isn't so easy to add as a simple method on the frame object without introducing the circular reference (not sure if that's OK to do).

joerick:
Hi @asmeurer ! I'm not sure how we'd incorporate visually into the output, but it should be easy enough to write a renderer to do it - the Frame objects do reference their parent and there's already a property : https://github.com/joerick/pyinstrument/blob/master/pyinstrument/frame.py#L21 that gives a number between 0 and 1 indicating the fraction of parent time.

joerick:
Percentage time was added in v4.3.0, although this is relative to the total, not the parent.

PR #174, opened 2022-01-05
Relative percentage display for Console output

Khreas:
Hello!

First of all, thanks for the great package. I discovered it recently but am already a big fan of it! 🎇

Issue #132 and issue #171 mentioned the fact that displaying percentages alongside actual timings could be interesting. The kind of percentages discussed in the two issues are not exactly the same. #171 mentioned the percentage of time a frame took to execute compared to **the whole execution time** while #132 mentioned the percentages of time a frame took **relatively to its direct siblings**.

Since I am also interested in this kind feature, I have implemented a specific processor that computes the percentage of time for a specific frame given its direct siblings. Switching to the overall execution percentage should be easy by modifying the said processor.

Plus, I modified the ConsoleRenderer and the Profiler classes by adding a parameter that controls whether those percentages should be displayed or not. Finally, I added a command line switch that would trigger the

Here's what it looks like:

image

I did not push the development too far (like adding that to the HTML output) for the moment to leave room for discussion, but if you think that such changes could be of any interest, I'll gladly add support to HTML / JSONRenderers!

Have a nice day 🚀

### DIFF
--- a/pyinstrument/__main__.py
+++ b/pyinstrument/__main__.py
@@ -162,6 +162,14 @@ def main():
         help="show everything",
         default=False,
     )
+    parser.add_option(
+        "",
+        "--show-percentage",
+        dest="show_percentage",
+        action="store_true",
+        help="(text renderer only) show relative percentages alongside timings",
+        default=False,
+    )
 
     parser.add_option(
         "",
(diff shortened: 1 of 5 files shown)


joerick:
Hi @Khreas, thanks for sending this over! I personally think that percentages relative to the total as seen in #171 might be more useful. Would that suit your use case also? Additionally, I wanted to have a look at how we might generalise the renderer options so that other renderers can add options without further expanding the command line interface. So I might take a crack at this myself and see where I end up.

PR #180, opened 2022-04-27
Add flag --filter-threshold

steinarvk-kolonial:
With the currently-hardcoded filter-threshold of 0.01,
some output including speedscope output can be quite misleading:
frames that are individually tiny but called often enough to
add up to a significant amount of time get eliminated.

This commit adds --filter-threshold as a flag, in particular
allowing --filter-threshold=0 which includes every frame no
matter how small.


steinarvk-kolonial:
Closing this because you seem to have an implementation ready at #183 .

PR #182, opened 2022-04-27
Flag --save-raw-report to save by filename

steinarvk-kolonial:
Instead of identifier for a file in report_dir().

This is easier if you're using pyinstrument in a non-interactive
context, e.g. you're running the profiler on machine A but
copying the file over to be analyzed on machine B.

### DIFF
--- a/pyinstrument/__main__.py
+++ b/pyinstrument/__main__.py
@@ -74,7 +74,21 @@ def main():
     )
 
     parser.add_option(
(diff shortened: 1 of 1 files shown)


joerick:
This is cool, but I am thinking it might be better as a Renderer, because I guess you're foregoing the normal output in this mode... `pyinstrument -r pyisession -o out.pyisession myscript.py`

That does mean that the renderer API needs to be a bit more general. I think that's okay. I'll take a look at that refactor.

PR #183, opened 2022-05-10
Add cmdline interface for renderer options and add percent_of_total option

joerick:
Adds a command-line option `-p` `--render-option` that allows arbitrary setting of render options.

For example, to show percentages on the command line, you can do `pyinstrument -p time=percent_of_total myscript.py`.

Processor options can also be set, using a JSON dict form, like `pyinstrument -p 'processor_options={"filter_threshold": 0}' myscript.py`, or dotted form, like `pyinstrument -p processor_options.filter_threshold=0 myscript.py`

Closes #180 and closes #174

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -14,7 +14,7 @@ repos:
         name: isort (python)
 
 -   repo: https://github.com/psf/black
-    rev: 20.8b1
+    rev: 22.3.0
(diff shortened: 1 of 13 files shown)


steinarvk-kolonial:
Seems like this would remove the need for https://github.com/joerick/pyinstrument/pull/180 !

PR #193, opened 2022-05-27
Add two more docs examples

tonybaloney:
Adds pytest (single test) and FastAPI using code I've deployed many times :-)

Closes #177

### DIFF
--- a/docs/guide.md
+++ b/docs/guide.md
@@ -247,8 +247,13 @@ To invoke, make any request to your application with the GET parameter `profile=
 
 ### Profile Pytest tests
 
-Pyinstrument can be invoked via the command-line to run pytest, but you will see a consolidated report for the test suite.
-To instrument specific tests, create and auto-use fixture in `conftest.py` in your test folder:
+Pyinstrument can be invoked via the command-line to run pytest, giving you a consolidated report for the test suite.
+
+```
+pyinstrument -m pytest [pytest-args...]
+```
+
+Or, to instrument specific tests, create and auto-use fixture in `conftest.py` in your test folder:
 
 ```python
 from pathlib import Path
--- a/docs/guide.md
+++ b/docs/guide.md
@@ -184,6 +184,96 @@ This will check for the `?profile` query param on each request and if found,
 it starts profiling. After each request where the profiler was running it
 creates the html output and returns that instead of the actual response.
 
+### Profile a web request in FastAPI
+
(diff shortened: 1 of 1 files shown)


tonybaloney:
Just saw #177 so this would be an alternative implementation to that suggestion.

tonybaloney:
Just saw #177 so this would be an alternative implementation to that suggestion.

docs/guide.md
```diff
@@ -184,6 +184,96 @@ This will check for the `?profile` query param on each request and if found,
...
+    async def profile_request(request: Request, call_next):
+        profiling = request.query_params.get("profile", False)
+        if profiling:
+            profile = FastApiProfiler(interval=settings.profiling_interval)
+            profile.start()
+            await call_next(request)
+            profile.stop()
+            return HTMLResponse(profile.get_result_html())
+        else:
+            return await call_next(request)
```
joerick, line 241: I wonder if the FastAPIProfiler class is needed at all? Any reason not to call straight to pyinstrument in this function?

```suggestion
    from pyinstrument import Profiler
    async def profile_request(request: Request, call_next):
        profiling = request.query_params.get("profile", False)
        if profiling:
            profiler = Profiler(interval=settings.profiling_interval, async_mode="enabled")
            profiler.start()
            await call_next(request)
            profiler.stop()
            return HTMLResponse(profiler.output_html())
        else:
            return await call_next(request)
```

docs/guide.md
```diff
@@ -184,6 +184,96 @@ This will check for the `?profile` query param on each request and if found,
...
+
+```
+
+To invoke, make any request to your application with the GET parameter `profile=1` and it will print the HTML result from pyinstrument.
+
+### Profile Pytest tests
+
+Pyinstrument can be invoked via the command-line to run pytest, but you will see a consolidated report for the test suite.
+To instrument specific tests, create and auto-use fixture in `conftest.py` in your test folder:
```
joerick, line 250:
~~~suggestion
Pyinstrument can be invoked via the command-line to run pytest, giving you a consolidated report for the test suite.

```
pyinstrument -m pytest [pytest-args...]
```

Or, to instrument specific tests, create and auto-use fixture in `conftest.py` in your test folder:
~~~

joerick (commented):
I'm no expert in FastAPI, so I'm very happy to take your word on it! :) Nice to have a pytest example, too.

Issue #188, opened 2022-05-12
Profiling pytest tests


felixriese:
I just figured out that `pyinstrument` also works well with `pytest` tests, really nice!

Should we add this to the documentation?
(If this is wanted, I can try to make a PR)

```bash
pyinstrument -m pytest test.py
```

joerick:
Yeah! Send over a PR :)

felixriese:
@joerick I added the PR #195 :)
(no idea if I needed a fork for that. hope that works for you)

joerick:
Fixed in #193. Thank you for the suggestion!

PR #195, opened 2022-05-30
Add pytest example

felixriese:
See #188


joerick:
Thanks @felixriese ! But we in fact already have a PR for pytest documentation at #193. I'll close this in favour of that.

felixriese:
Even better, thanks! Did not see this PR :)

PR #199, opened 2022-06-20
3.11 frame support

### LINKED ISSUE #187: Python 3.11 compatibility
I have a project that uses pyinstrument as a dev dependency and it's failing to build on Python 3.11: https://github.com/adriangb/di/runs/6377204976?check_suite_focus=true#step:5:156

I can confirm this also happens locally with a simple `pip install pyinstrument` on 3.11.

tonybaloney:
Fixes #187 

Fork of the 3.11 testing branch.

Adds a shim for getting f_back.

In 3.11 this can be fetched using a function. See https://docs.python.org/3.11/whatsnew/3.11.html#pyframeobject-3-11-hiding

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -26,7 +26,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, windows-latest, macos-latest]
-        python-version: ["3.7", "3.8", "3.9", "3.10"]
+        python-version: ["3.7", "3.8", "3.9", "3.10", "3.11-dev"]
 
     steps:
     - uses: actions/checkout@v2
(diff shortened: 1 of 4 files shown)


joerick:
Thank you Anthony! I've been meaning to get around to this. I'm pleased that the modifications are limited to f_back, I had feared it might be worse. 

Not sure what's going on with CI, I've restarted it.

tonybaloney:
Looking at the implementation of PyFrame_GetBack, it returns a new reference, whereas frame->f_back is using a borrowed reference. This PR should probably decrement that reference once its completed, although frame objects are destroyed by the interpreter anyway..

joerick (approved):

joerick:
Thanks @tonybaloney! I'll get a release out now.

PR #203, opened 2022-07-15
Refactor to track information other than just time

joerick:
Supercedes #166.

This refactor allows pyinstrument to collect information other than the code object that was executing during the profiling. Currently, this function is only used for the class name, but I've got plans to capture other information.

Because the extra information ('attributes') isn't part of the identifier, it can vary between invocations. This was important for the class name, where the 'self' variable could change at runtime, while the same frame was executing. But this will also allow tracking of the currently executing line number, which unlocks the ability for pyinstrument to have line-by-line profiling, as well as tree-style profiling.


joerick:
I have a 3.11 version working, but I'm not sure about it... see this [forum post](https://discuss.python.org/t/getting-the-class-name-of-a-code-frame-object-in-cpython-3-11-c-api/17396).

joerick:
The 3.11 support is firmed-up. I think I'm ready to go with this. I might just do a little extra testing to make sure there are no perf regressions.

joerick:
All good! Merging...

joerick:
Just a note, this is merged into master but wasn't part of the 4.3.0 release, I wanted to hold this feature since it _might_ warrant a 5.0.0 release...

Issue #169, opened 2021-12-10
High memory consumption


Andrei-Aksionov:
I am using pyinstrument for profiling some ML/NLU specific tasks and noticed that when running profiling on function with a lot of simple computations inside memory consumption stedealy increases.

For example such a dummy code:
```python
for _ in range(10_000_000):
    res = [x for x in ["qwerty"] * 100]
```
when running `python main.py` consumes ~3.2 MB, while `pyinstrument main.py` - ~40 MB.
And the main issue is that while running without profiling memory consumption stays at the same level, with profiling - constantly increases.

Andrei-Aksionov:
Thanks for the response.

`interval` parameter affects memory consumption for long running program quite a bit.
I think such information should be reflected in the README file (not as Known issues, but something like good to know).
In docs I noticed only [this](https://pyinstrument.readthedocs.io/en/latest/guide.html?highlight=interval#profile-a-specific-chunk-of-code) mentioning regarding interval value, but as for me:

> but keep in mind that smaller intervals could affect the **performance overhead of profiling**.

means that the code will be running slower with smaller interval value, not that memory will be consumed like crazy.

In addition, as I understand, it's not possible to set the interval value from CLI.

Plus, it's not a solution, more like a workaround. Memory consumption still increases only now with a slower pace (with bigger interval value).
Maybe in some future you are planning to change  the code so that it dumps profiling data to disk in chunks in order to not store everything in memory and before preparing output to load all previously dumped data?

dralley:
> Maybe in some future you are planning to change the code so that it dumps profiling data to disk in chunks in order to not store everything in memory and before preparing output to load all previously dumped data?

This would be great.  We have some jobs which take 20 minutes, and by that point the memory overhead of pyinstrument grows to >20gb.  Without pyinstrument it's under 5gb.  On some systems this causes out of memory errors.

joerick:
Thanks for chiming in. I'm curious if reducing the profiling interval would be a solution for your use case, or if you really need millisecond accuracy. The issue I see with dumping intermediate output to disk is that at some point pyinstrument is going to have to load it all into RAM to analyse it and produce reports, which is going to hit the same RAM limits as you're seeing during execution.

dralley:
I don't suppose there is any way to compress or filter out unnecessary data during the runtime?

joerick:
Compress, maybe, but we'd still have the issue when analysing. What would you consider unnecessary data?

[...]

PR #207, opened 2022-08-09
Expose interval option in command-line interface

matt-graham:
Related to the discussion in #169, I've found that the memory overhead of using `pyinstrument` to profile long-running (several hours run time) scripts using the command-line interface with the default sampling interval of 0.001&thinsp;s can result in runs prematurely terminating with an out-of-memory error. Given there will be (tens of) millions of stack samples being recorded in this case, that there are issues with memory consumption in such cases is not surprising, but also this suggests a much longer sampling interval would be appropriate in such cases while still giving reasonable low-variance estimates. This PR exposes the existing `interval` argument to the `Profiler` class in the command-line interface to allow easier adjustment of the sampling interval without needing to use the Python interface directly.

### DIFF
--- a/pyinstrument/__main__.py
+++ b/pyinstrument/__main__.py
@@ -218,6 +218,20 @@ def main():
         action="store_false",
         help="(text renderer only) force no color text output",
     )
+    parser.add_option(
+        "-i",
+        "--interval",
+        action="store",
+        type=float,
+        help=(
+            "Minimum time, in seconds, between each stack sample. Smaller values "
+            "allow resolving shorter duration function calls but conversely incur a "
+            "greater runtime and memory consumption overhead. For longer running "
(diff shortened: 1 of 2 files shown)


matt-graham:
Not sure if [the failing `test_interval` test on `macos-latest, 3.11-dev`](https://github.com/joerick/pyinstrument/runs/7744548124?check_suite_focus=true) was due to the larger interval in `test_interval` (0.005s) compared to the default (0.001s) meaning there is a non-negligible probability of some of the function calls not being captured and its just chance this was only on the `macos-latest, 3.11-dev` run, or there is something more specific going on here. I have tried reducing the interval down to 0.002s to see if that helps and also fixed the missing new line at end of file causing the pre-commit fail.

PR #217, opened 2022-09-27
skip hidden frames

### LINKED ISSUE #190: Proposal: Optionally Skip Frames in Output

telamonian:
fixes #190

updated version of #211. It was easier to just start a new PR than it was to resolve the conflicts after the py 3.11 updates. From the original PR:

> This adds a generic "hidden" property to the frame record objects. Currently this is only set to true if the frame has __tracebackhide__==True in it's locals dict, as discussed in https://github.com/joerick/pyinstrument/issues/190. Also adds a default processor that hides any frames with .hidden==True.
>
> Not sure if this is the right approach overall, and very not sure if the changes to the c extension were the right ones. @joerick Feedback would be much appreciated.
>
> I also added the .devcontainer dir that I came up with while developing these changes. This allows for developing pyinstrument in a container when using vscode (and maybe other ides?). Specifically, this allowed me to do stuff like simultaneously debug the python and c extension code (even though I'm developing on a mac), which was super useful. Dunno if you want this in the repo, I can remove it if you prefer.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,0 +1,23 @@
+# See here for image contents: https://github.com/microsoft/vscode-dev-containers/tree/v0.238.0/containers/cpp/.devcontainer/base.Dockerfile
+
+# [Choice] Debian / Ubuntu version (use Debian 11, Ubuntu 18.04/22.04 on local arm64/Apple Silicon): debian-11, debian-10, ubuntu-22.04, ubuntu-20.04, ubuntu-18.04
+ARG VARIANT="bullseye"
+FROM mcr.microsoft.com/vscode/devcontainers/cpp:0-${VARIANT}
+
+# [Optional] Install CMake version different from what base image has already installed. 
+# CMake reinstall choices: none, 3.21.5, 3.22.2, or versions from https://cmake.org/download/
+ARG REINSTALL_CMAKE_VERSION_FROM_SOURCE="none"
+
+# Optionally install the cmake for vcpkg
+COPY ./reinstall-cmake.sh /tmp/
+RUN if [ "${REINSTALL_CMAKE_VERSION_FROM_SOURCE}" != "none" ]; then \
+        chmod +x /tmp/reinstall-cmake.sh && /tmp/reinstall-cmake.sh ${REINSTALL_CMAKE_VERSION_FROM_SOURCE}; \
+    fi \
+    && rm -f /tmp/reinstall-cmake.sh
+
+# [Optional] Uncomment this section to install additional vcpkg ports.
+# RUN su vscode -c "${VCPKG_ROOT}/vcpkg install <your-port-name-here>"
+
+# [Optional] Uncomment this section to install additional packages.
+# RUN apt-get update && export DEBIAN_FRONTEND=noninteractive \
+#     && apt-get -y install --no-install-recommends <your-package-list-here>
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -4,7 +4,7 @@
 ARG VARIANT="bullseye"
 FROM mcr.microsoft.com/vscode/devcontainers/cpp:0-${VARIANT}
 
-# [Optional] Install CMake version different from what base image has already installed. 
+# [Optional] Install CMake version different from what base image has already installed.
 # CMake reinstall choices: none, 3.21.5, 3.22.2, or versions from https://cmake.org/download/
 ARG REINSTALL_CMAKE_VERSION_FROM_SOURCE="none"
 
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,46 @@
+// For format details, see https://aka.ms/devcontainer.json. For config options, see the README at:
+// https://github.com/microsoft/vscode-dev-containers/tree/v0.238.0/containers/cpp
+{
+	"name": "C++",
+	"build": {
+		"dockerfile": "Dockerfile",
+		// Update 'VARIANT' to pick an Debian / Ubuntu OS version: debian-11, debian-10, ubuntu-22.04, ubuntu-20.04, ubuntu-18.04
+		// Use Debian 11, Ubuntu 18.04 or Ubuntu 22.04 on local arm64/Apple Silicon
+		"args": { "VARIANT": "ubuntu-22.04" }
+	},
+	"runArgs": ["--cap-add=SYS_PTRACE", "--security-opt", "seccomp=unconfined"],
+
+	// Configure tool-specific properties.
+	"customizations": {
+		// Configure properties specific to VS Code.
+		"vscode": {
+			// Add the IDs of extensions you want installed when the container is created.
+			"extensions": [
+				"benjamin-simmonds.pythoncpp-debug",
+				"eamodio.gitlens",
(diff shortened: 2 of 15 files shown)


telamonian:
I think I've gotten the c api code correct, so this should be ready for your eyeballs now @joerick. One specific question I had about my code:

https://github.com/joerick/pyinstrument/blob/96831856d1d7dfa167fb634b89fa42d982004f40/pyinstrument/low_level/stat_profile.c#L490-L491

In the meantime, I'll try to add a test function a la:

https://github.com/joerick/pyinstrument/blob/eae9e14df7498ee93beb553254ab0d7480de527a/test/test_processors.py#L28-L79

telamonian:
i just fixed up the code defining the `Frame.hidden` property on the python side, but that raised another question. @joerick Why is this a property of `Frame`:

https://github.com/joerick/pyinstrument/blob/eae9e14df7498ee93beb553254ab0d7480de527a/pyinstrument/frame.py#L139-L142

but not something like:

```python
    @property
    def line_number(self) -> int | None:
        l = self.get_attribute_value(ATTRIBUTE_MARKER_LINE_NUMBER)

        return None if l is None else int(l)
```

(ie why is `ATTRIBUTE_MARKER_LINE_NUMBER` defined but not used?). This probably isn't super important, but I just want to make sure that I'm handling `Frame.hidden` correctly

telamonian:
I added/expanded some unittests to cover the low level behavior for `__tracebackhide__==True` and the high level`remove_hidden_frames` processor. However, I ran into a seemingly tricky issue involving updates to the locals dict on the C side. The following version of `test_frame_info.py` fails both for my branch and main:

```python
import inspect

from pyinstrument.low_level import stat_profile as stat_profile_c
from pyinstrument.low_level import stat_profile_python


class AClass:
    def get_frame_info_with_cell_variable(self, getter_function):
        def an_inner_function():
            __tracebackhide__ = True

            frame = inspect.currentframe()
            assert frame

            # reference self to make it a cell variable
            if self:
                pass

            return frame

        frame = an_inner_function()
        return getter_function(frame)


def test_frame_info_with_classes():
    instance = AClass()

    test_functions = [
        instance.get_frame_info_with_cell_variable,
    ]

    for test_function in test_functions:
        c_frame_info = test_function(stat_profile_c.get_frame_info)
        py_frame_info = test_function(stat_profile_python.get_frame_info)

        assert c_frame_info == py_frame_info
```

On main branch the failure is:

```
>           assert c_frame_info == py_frame_info
E           AssertionError: assert 'an_inner_fun...y\x009\x01l19' == 'an_inner_fun...AClass\x01l19'
E             Skipping 62 identical leading characters in diff, use -v to show
E             - _foo.py9cAClassl19
E             ?           --------
E             + _foo.py9l19
```

on my branch the failure is:

```
>           assert c_frame_info == py_frame_info
E           AssertionError: assert 'an_inner_fun...\x01l19\x01h0' == 'an_inner_fun...\x01l19\x01h1'
E             Skipping 62 identical leading characters in diff, use -v to show
E             - _foo.py9cAClassl19h1
E             ?           --------     ^
E             + _foo.py9l19h0
E             ?                ^
```

@joerick Any ideas?

Not sure how significant the failure of the class attribute showing up in the c frame info string in this test is. It's possible that in the normal functioning of pyinstrument the locals dict gets "rung" enough on the c side for this not to happen anyway.

Also note I'm still dev-ing everything in py 3.10. I'll test this out in py 3.11 later

telamonian:
i did some simultaneous python/c debugging (using my fancy-schmancy devcontainer, which is neat) and got to the bottom of the inconsistency in how the c func was marking frames as hidden. Basically, one of my sanity checks at the start of the `_get_tracebackhidden` func was malformed and checking the wrong thing.

@joerick I fixed that, and updated the unittests in `test_frame_info.py` to more thoroughly cover the various cases related to the behavior of `__tracebackhide__`. I'm now fairly confident that all of the code related to hiding frames in response to `__tracebackhide__` is correct, both on the c and python sides. However, I am still getting basically the same error related to the class name attribute as I pointed out in the previous message above:

```
>           assert c_frame_info == py_frame_info
E           AssertionError: assert 'an_inner_fun...\x01l26\x01h1' == 'an_inner_fun...\x01l26\x01h1'
E             Skipping 70 identical leading characters in diff, use -v to show
E             - nfo.py16cAClassl26h1
E             ?           --------
E             + nfo.py16l26h1

test/low_level/test_frame_info.py:60: AssertionError
```

It's got something to do with `self` being treated as a cellvar in the nested func (i think. I found some [related docs here](https://docs.python.org/3/reference/datamodel.html#the-standard-type-hierarchy)). How should I proceed?

joerick:
> i just fixed up the code defining the `Frame.hidden` property on the python side, but that raised another question. @joerick Why is this a property of `Frame`:

Yeah, that's a little confusing. The reason is that the 'line number' refers to the code object of that frame - e.g. the line number of the first line of the function. That uniquely identifies the function, and remains static throughout the execution, so is part of the identifier. But the line number _attribute_ is runtime information - it changes as the program executes. 

The runtime attribute isn't used anywhere in the codebase yet, I'm currently working on a branch that uses it to output line-by-line information. Hopefully that will work soon!

joerick:
> It's got something to do with `self` being treated as a cellvar in the nested func (i think. I found some [related docs here](https://docs.python.org/3/reference/datamodel.html#the-standard-type-hierarchy)). How should I proceed?

That's an interesting one. In a way, the C version is correct - the inner function isn't a method on a class. The difference is that now you've defined the hidden property, I'm guessing the first local is your local `__tracebackhide__` var, and not the `self` variable.

Clearly, it's not an issue for your code. I'd remove the `__tracebackhide__` from the `test_frame_info_with_classes` test. Maybe there's a better way to test the cell self variable, I might revisit that one day.

BTW, it would also be good to have a frame_info test case where the `__tracebackhide__` variable isn't set, just so we test the opposite too.

telamonian:
> I'd remove the `__tracebackhide__` from the `test_frame_info_with_classes` test. Maybe there's a better way to test the cell self variable, I might revisit that one day.
> 
> BTW, it would also be good to have a frame_info test case where the `__tracebackhide__` variable isn't set, just so we test the opposite too.

@joerick done

For the sake of clarity 6 months from now when someone looks at this bit of code again, I also added another test case with `__tracebackhide__ == False` and a note about how it should behave the same as the `True` case.

telamonian:
@joerick What is your thinking on this PR? Is there anything else that needs to be done before it can be pulled in?

joerick:
I've just done a pass on this, I think it should be good-to-go.

joerick:
Thanks @telamonian ! I'll see about a release.

PR #221, opened 2022-11-10
Add __all__ for type hint

kigawas:


joerick:
Thanks! Can you add `__version__` to that too?

joerick:
Oh, and `load_ipython_extension` will be needed, I expect.

PR #224, opened 2022-11-22
Adds falcon example using middleware

abhinand-c:
Adds an example to docs and in example folder for integrating with falcon web framework using middleware.


abhinand-c:
Updated branch to clear unused code.

PR #236, opened 2023-04-09
ENH: add a renderer to save python-profile prof files

### LINKED ISSUE #64: pstats output
One thing that would be very useful is to be able to produce output from pyinstrument in a Python pstats file. That would allow it to, for example, be passed to `gprof2dot` to produce a call-graph tree. It looks like the file format is pretty simple -- some data in a dictionary gets dumped using marshal.dump (see [here](https://github.com/python/cpython/blob/master/Lib/pstats.py#L179)).

mattip:
Closes #64. Now running `python -m pyinstrument -r prof -o <filename> <script>` will create a file compatible with `python -m profile -o <filename> <script>`. It can be loaded into tools like `gprof2dot` and `snakeviz` for visualization.

Of course the `ncall` number is bogus. I used -1. Since we only know total time, the `percall` time is also not available. Ideas for further tests or correcting the statistics are welcome.


pyinstrument/renderers/profrenderer.py
```diff
@@ -0,0 +1,64 @@
...
+        if frame is None:
+            return {}
+
+        key = (frame.file_path or "", frame.line_no or 0, frame.function)
+        val = [-1, -1, frame.time, frame.total_self_time, {}]
+        if frame.parent:
+            p = frame.parent
+            val[-1][(p.file_path or "", p.line_no or 0, p.function)] = val[:-1]
+        if key in stats:
+            return
```
mattip, line 35: The code in lines 36-38 below is not used. I thought there would be a way to collect better statistics, but none came to mind.

pyinstrument/__main__.py
```diff
@@ -456,15 +461,17 @@ def get_renderer_class(renderer: str) -> type[renderers.Renderer]:
...
         try:
             return object_with_import_path(renderer)
         except (ValueError, ModuleNotFoundError, AttributeError) as err:
             # ValueError means we failed to import this object
             raise OptionsParseError(
                 f"Failed to find renderer with name {renderer!r}.\n"
-                "Options are text, html, json, speedscope or a Python import path to a Renderer\n"
-                "class.\n"
+                "Options are text, html, json, speedscope, prof or a Python\n"
+                "import path to a Rendererclass.\n"
```
joerick, line 474:
```suggestion
                "import path to a Renderer class.\n"
```

pyinstrument/renderers/profrenderer.py
```diff
@@ -0,0 +1,64 @@
...
+
+    def __init__(self, **kwargs: Any):
+        super().__init__(**kwargs)
+
+    def render_frame(self, frame: Frame | None, stats: dict) -> dict:
+        if frame is None:
+            return {}
+
+        key = (frame.file_path or "", frame.line_no or 0, frame.function)
+        val = [-1, -1, frame.time, frame.total_self_time, {}]
```
joerick, line 30:
```suggestion
        # format is (call_time, number_calls, total_time, cumulative_time, callers)
        val = [-1, -1, frame.time, frame.total_self_time, {}]
```

pyinstrument/renderers/profrenderer.py
joerick: would you mind removing them if so?

joerick (commented):
Hey Matti! Thanks for this, it looks good. Would it be possible to add just a test that checks that [`pstats.Stats`](https://docs.python.org/3/library/profile.html#module-pstats) doesn't crash when loading the output of a simple script?

pyinstrument/__main__.py
mattip: done

pyinstrument/renderers/profrenderer.py
mattip: done

mattip:
I responded to the review and added a test

mattip:
I think this is ready for another round of review. Perhaps enabling the CI run would give some confidence that the tests are correct?

joerick:
Thanks and apologies for the delay! I'll try to look in the next few days

pyinstrument/__main__.py
```diff
@@ -325,7 +325,12 @@ def dash_m_callback(option: str, opt: str, value: str, parser: optparse.OptionPa
         session = profiler.stop()
 
     if options.outfile:
-        f = codecs.open(options.outfile, "w", "utf-8")
+        if should_close_f_after_writing:
+            f.close()
```
joerick, line 329: This seems like it's in the wrong place, perhaps an accidental addition? I think these lines should be removed.

```suggestion
```

joerick: Actually, there's something weird going on here... the outfile is opened twice. Must be an existing defect. I'll take a look.

joerick (commented):
This is complicated by the fact that the prof renderer outputs bytes, not strs. I'll have a think how we can resolve this neatly.

pyinstrument/__main__.py
joerick: I've refactored this to remove the preexisting issue, I'll push the changes

pyinstrument/renderers/profrenderer.py
```diff
@@ -0,0 +1,60 @@
...
+            return {}
+
+        key = (frame.file_path or "", frame.line_no or 0, frame.function)
+        # Format is (call_time, number_calls, total_time, cumulative_time, callers)
+        val = [-1, -1, frame.time, frame.total_self_time, {}]
+        if frame.parent:
+            p = frame.parent
+            val[-1][(p.file_path or "", p.line_no or 0, p.function)] = val[:-1]
+        if key in stats:
+            return
```
joerick: I think this is an issue - we shouldn't discard frames when they've already appeared in stats, we should combine their data. so times should accumulate. I think I have a fix for this, will push shortly.

joerick:
I've pushed some fixes to this, if you get a sec could you check this is still working for you @mattip? There were a couple issues with the prof rendering as well as some internal issues (like total_self_time always returning zero), but I think I got them all.

pyinstrument/renderers/profrenderer.py
```diff
@@ -0,0 +1,84 @@
...
+            number_calls = -1
+            total_time = 0
+            cumulative_time = 0
+            callers: dict[FrameKey, CallerValue] = {}
+        else:
+            call_time, number_calls, total_time, cumulative_time, callers = stats[key]
+
+        # update the total time and cumulative time
+        total_time += frame.total_self_time
+        cumulative_time += frame.time
```
mattip, line 53: Nice, thanks for fixing that.

mattip:
I think this looks good on a toy script I tried. Is there any end-to-end testing you do offline to verify that the renderers all give similar results?

joerick:
Normally I just play with some of the examples scripts in /examples. I checked these with snakeviz. But it would be good to double check the callers dict somehow, maybe there's some way to visualise the call graph? I seem to remember some tool that had blobs that represent functions and lines representing calls...The other thing we could do is assert more deeply in a unit test, but I'm not sure I'm confident enough how the data should look...! I suppose one could compare a toy script with cProfile to verify. Sent from my phoneOn 26 Apr 2023, at 03:38, Matti Picus ***@***.***> wrote:﻿
I think this looks good on a toy script I tried. Is there any end-to-end testing you do offline to verify that the renderers all give similar results?

mattip:
> maybe there's some way to visualise the call graph

Running the code in the new test, and then using `gprof2dot`
```
gprof2dot -f pstats /tmp/test_renderer.prof | dot -Tpng -o /tmp/output.png
```

I get 
output

mattip:
Sorry, that was cProfile. The ProfRenderer from pyinstrument gives me this. Note the `50% -1` in the call to `e` and `sleep`. I wonder what is wrong

output

joerick:
nice, thank you! well the "-1" is because the number of calls is -1, due to statistical profiling we don't know that. but the 50% below the branches is odd, and looks wrong. Perhaps it is getting confused with the numcalls being -1? we could try 1 to see if that helps. Otherwise it's probably something to do with the callers dict

mattip:
Yes, that did fix the per-cent. The rendering is a bit off, it is hard to read 100%. But that is certainly not pyinstrument's problem :). I think this is about as good as we can do?
output

joerick:
yeah that looks good to me. I'd rather the numcalls said -1 rather than 1 or 0 as it's clear that there is no data.

I think i'd like some more detailed unit tests on the things we've discovered during this PR about the pstats file format. mostly, just basic stuff around frames, timings, and asserting that the fix we made above still holds. If you have time, that would be great, otherwise I'll get around to it at some point.

Something else I'd wondering - is `pstats` a better name for this than `prof`?

mattip:
> is pstats a better name for this than prof?

yes, I am horrible at naming things. I will take a look at adding tests. Since the `stats` output is a dictionary, I think it should be straightforward to assert some of what you suggest.

mattip:
Added some tests and renamed the renderer.

mattip:
Rebased off main, and CI is passing.

joerick (commented):
Looking over the diff, still a few instances of the renderer referred to as 'prof' and not pstats

mattip:
Good catch. I changed the file extension for this renderer to `*.pstats`.

joerick (approved):

joerick:
Looks good! Thanks @mattip !

mattip:
Thanks for shepherding this through to the end.

mattip:
Is there a plan for a new release?

joerick:
Sorry for the wait, gonna get #240 in and then cut a new release

mattip:
Cool, thanks

PR #239, opened 2023-04-30
Fix show all

mattip:
Replaces #229

Builds on the previous commits. I reverted some changed files according to review comments, and  linted the changes made. I also rebased off `main`


joerick:
excellent, thank you @mattip !

PR #240, opened 2023-05-03
Add flat mode to console renderer

matthiasdiener:
This was inspired by a similar flat mode in [vmprof](https://github.com/vmprof/vmprof-python#flattened-output). Sometimes it is interesting to focus on functions that get called from multiple locations in the call graph in order to determine which functions consume the most time overall.

Run with `pyinstrument ... -p flat`.

Here is how it looks like:

```console
$ pyinstrument --load-prev 2023-05-03T14-25-05  # standard mode

  _     ._   __/__   _ _  _  _ _/_   Recorded: 14:25:05  Samples:  165
 /_//_/// /_\ / //_// / //_'/ //     Duration: 1.285     CPU time: 0.068
/   _/                      v4.4.0

Program: examples/c_sort.py

1.284 <module>  ../examples/c_sort.py:1
├─ 1.193 slow_key  ../examples/c_sort.py:27
│  └─ 1.193 sleep  None
│        [2 frames hidden]  ..
└─ 0.090 <module>  numpy/__init__.py:1
      [158 frames hidden]  numpy, pickle, .., _compat_pickle, __...

    pyinstrument --load-prev 2023-05-03T14-25-05 [options]

$ pyinstrument --load-prev 2023-05-03T14-25-05 -p flat  # flat mode

  _     ._   __/__   _ _  _  _ _/_   Recorded: 14:25:05  Samples:  165
 /_//_/// /_\ / //_// / //_'/ //     Duration: 1.285     CPU time: 0.068
/   _/                      v4.4.0

Program: examples/c_sort.py

1.193s sleep  ../<built-in>:0
0.023s <module>  numpy/random/_pickle.py:1
0.018s <module>  numpy/core/__init__.py:1
0.018s <module>  numpy/core/overrides.py:1
0.018s <module>  numpy/__init__.py:1
0.013s <module>  numpy/lib/__init__.py:1
0.000s <module>  ../examples/c_sort.py:1
0.000s slow_key  ../examples/c_sort.py:27
0.000s <module>  numpy/core/multiarray.py:1
0.000s <module>  numpy/random/__init__.py:1

    pyinstrument --load-prev 2023-05-03T14-25-05 [options]

### DIFF
--- a/pyinstrument/renderers/console.py
+++ b/pyinstrument/renderers/console.py
@@ -1,5 +1,6 @@
 import time
-from typing import Any
+from dataclasses import dataclass
+from typing import Any, Dict, List, Optional, Tuple
 
 import pyinstrument
 from pyinstrument import processors
@@ -22,6 +23,7 @@ class ConsoleRenderer(FrameRenderer):
         self,
         unicode: bool = False,
         color: bool = False,
+        flat: bool = False,
         time: LiteralStr["seconds", "percent_of_total"] = "seconds",
         **kwargs: Any,
     ):
@@ -34,6 +36,7 @@ class ConsoleRenderer(FrameRenderer):
 
         self.unicode = unicode
         self.color = color
+        self.flat = flat
         self.colors = self.colors_enabled if color else self.colors_disabled
         self.time = time
 
@@ -48,7 +51,10 @@ class ConsoleRenderer(FrameRenderer):
 
         self.root_frame = frame
 
-        result += self.render_frame(self.root_frame)
+        if self.flat:
+            result += self.render_frame_flat(self.root_frame)
+        else:
+            result += self.render_frame(self.root_frame)
         result += "\n"
 
         return result
@@ -83,10 +89,14 @@ class ConsoleRenderer(FrameRenderer):
             or frame in frame.group.exit_frames
         ):
             if self.time == "percent_of_total":
-                percent = self.frame_proportion_of_total_time(frame) * 100
-                time_str = self._ansi_color_for_time(frame) + f"{percent:.0f}%" + self.colors.end
+                percent = self.frame_proportion_of_total_time(frame.time) * 100
+                time_str = (
+                    self._ansi_color_for_time(frame.time) + f"{percent:.0f}%" + self.colors.end
+                )
             else:
-                time_str = self._ansi_color_for_time(frame) + f"{frame.time:.3f}" + self.colors.end
+                time_str = (
+                    self._ansi_color_for_time(frame.time) + f"{frame.time:.3f}" + self.colors.end
+                )
 
             name_color = self._ansi_color_for_name(frame)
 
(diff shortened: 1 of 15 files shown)

### CODE REVIEW

**Comment by matthiasdiener** [comment]
> Nice, thanks! I'll take a look soon. Just a quick note, 'exclusive time' is known as 'self time' in this project. I think it's already available as total_self_time on Frame. But, there may have been a bug in that property until recently, can't remember if it's merged or not!

Thanks! I tried this with `total_self_time`, but it always returned zero. Looking at the other PRs, I think this might be a fix: https://github.com/joerick/pyinstrument/pull/236/files#diff-34c9ac7ed6042dfd8c5340562bb37088e871be076ce870b1325da6e3f8f4d5cb

**Comment by matthiasdiener** [comment]
> > Nice, thanks! I'll take a look soon. Just a quick note, 'exclusive time' is known as 'self time' in this project. I think it's already available as total_self_time on Frame. But, there may have been a bug in that property until recently, can't remember if it's merged or not!
> 
> Thanks! I tried this with `total_self_time`, but it always returned zero. Looking at the other PRs, I think this might be a fix: https://github.com/joerick/pyinstrument/pull/236/files#diff-34c9ac7ed6042dfd8c5340562bb37088e871be076ce870b1325da6e3f8f4d5cb

I added that fix to this PR and modified the code to use `total_self_time`. Could you please approve the CI workflow?

File: pyinstrument/renderers/console.py [hunk 1]
```diff
@@ -137,11 +147,64 @@ def render_frame(self, frame: Frame, indent: str = "", child_indent: str = "") -
 
         return result
 
-    def frame_proportion_of_total_time(self, frame: Frame):
-        return frame.time / self.root_frame.time
+    def render_frame_flat(self, frame: Frame) -> str:
+        @dataclass(frozen=True, eq=True)
+        class FrameDesc:
+            code_position_short: Optional[str]
+            function: str
```
joerick, line 154: `frame.identifier` should do the job of uniquely identifying a frame's code location for the purpose of this computation, I don't think we'd need this dataclass.

pyinstrument/renderers/console.py
```diff
@@ -137,11 +147,64 @@ def render_frame(self, frame: Frame, indent: str = "", child_indent: str = "") -
...
+            res += "{color}{val:.3f}{unit}{c.end} {name_color}{function}{c.end}  {c.faint}{code_position}{c.end} {ncalls}\n".format(
+                color=color,
+                val=val,
+                unit=unit,
+                c=self.colors,
+                name_color=self._ansi_color_for_name(frame_desc_to_frame[frame_desc]),
+                function=frame_desc.function,
+                code_position=frame_desc.code_position_short,
+                ncalls=frame_desc_to_ncalls[frame_desc],
+            )
```
joerick, line 199: I'm wondering if there's any potential for code reuse of this bit. perhaps we can factor out a bit of  `render_frame` above - as an example, I note that this code won't print class names of functions... i think that there might be a potential for a `frame_description(self, frame: Frame, time: float) -> str` that captures all the text display stuff

joerick (commented):
Looks good @matthiasdiener, thank you! A few comments inline. The last commit you added isn't correct though, it's not possible to say anything about 'ncalls' in pyinstrument because it is a statistical profiler - they could be many calls that pyinstrument didn't record because it's only recording every {interval}s.

pyinstrument/renderers/console.py
matthiasdiener: Thanks! I believe d52a191 addresses this.

matthiasdiener:
> The last commit you added isn't correct though, it's not possible to say anything about 'ncalls' in pyinstrument because it is a statistical profiler - they could be many calls that pyinstrument didn't record because it's only recording every {interval}s.

You're right, of course. I reverted that commit.

pyinstrument/renderers/console.py
matthiasdiener: That's a good idea, thanks! I added `frame_description()` in [1171f3e](https://github.com/joerick/pyinstrument/pull/240/commits/1171f3e8d8f8e7b503e34d78d9a7e2ec19a898d7)

joerick:
leave this with me @matthiasdiener, I have some more thoughts on code structure that are probably easier to do than describe!

joerick:
Would you mind checking this still works as you intended @matthiasdiener ? I think I'm happy with it.

pyinstrument/renderers/console.py
```diff
@@ -22,22 +24,25 @@ def __init__(
...
+        flat: bool = False,
         time: LiteralStr["seconds", "percent_of_total"] = "seconds",
         **kwargs: Any,
-    ):
+    ) -> None:
         """
         :param unicode: Use unicode, like box-drawing characters in the output.
         :param color: Enable color support, using ANSI color sequences.
         :param time: How to display the duration of each frame - ``'seconds'`` or ``'percent_of_total'``
+        :param flat: Display a flat profile instead of a call graph.
```
matthiasdiener, line 35: I'd suggest keeping them in the same order as in the `__init__` parameters.

```suggestion
        :param flat: Display a flat profile instead of a call graph.
        :param time: How to display the duration of each frame - ``'seconds'`` or ``'percent_of_total'``
```

pyinstrument/renderers/console.py
```diff
@@ -137,11 +124,65 @@ def render_frame(self, frame: Frame, indent: str = "", child_indent: str = "") -
...
+        frame_id_to_frame: Dict[str, Frame] = {}
+
+        walk(frame)
+
+        id_time_pairs: List[Tuple[str, float]] = sorted(
+            frame_id_to_time.items(), key=(lambda item: item[1]), reverse=True
+        )
+
+        # remove nodes that represent less than 0.1% of the total time
+        id_time_pairs = [pair for pair in id_time_pairs if pair[1] / self.root_frame.time > 0.001]
```
matthiasdiener, line 148: Should this be affected by `--show-all`?

matthiasdiener (commented):
Thanks @joerick! This still works as intended.

pyinstrument/renderers/console.py
joerick: yeah good catch! i'll chuck that in too

matthiasdiener:
Thanks @joerick! This is still working as intended.

Issue #241, opened 2023-05-03
Incorrect Session.program when running a module


matthiasdiener:
Due to `sys.argv` modifications (in `__main__.py`?), the program name is stored incorrectly in `Session.program`:

actual output:
```console
$ pyinstrument -m mpi4py t.py -a 
Program: mpi4py -a
```

expected output:
```console
$ pyinstrument -m mpi4py t.py -a 
Program: mpi4py t.py -a
```

joerick:
I _think_ the issue here is that changing sys.argv is pretty common in these styles of wrapper scripts. pyinstrument, [mpi4py](https://github.com/mpi4py/mpi4py/blob/master/src/mpi4py/run.py#L31), cProfile all do it. Each time it happens, each script doesn't reset the value once the script has finished running, because each assumes it's the top-level script.

My thought is that we should consider changing sys.argv a 'patch' - i.e. it should be reset to it's previous value once the code has finished. I'll push a PR of a fix that I think works, and _should_ make more sense.

PR #243, opened 2023-05-19
session: show full command line using psutil

matthiasdiener:
What do you think of this @joerick ? This does require `psutils`, but has the advantage of working well on all Unix/Win platforms.

Addresses #241.

```console
$ pyinstrument -m mpi4py t.py --myargs
Program: /Users/mdiener/Work/emirge/miniforge3/envs/ceesd/bin/python3.11 /Users/mdiener/Work/emirge/miniforge3/envs/ceesd/bin/pyinstrument -m mpi4py t.py --myargs

### DIFF
--- a/pyinstrument/profiler.py
+++ b/pyinstrument/profiler.py
@@ -152,12 +152,16 @@ class Profiler:
 
         cpu_time = process_time() - self._active_session.start_process_time
 
+        import psutil
+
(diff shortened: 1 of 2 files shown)


joerick:
Hi @matthiasdiener ! Sorry it's taken so long for me to get to this, it took a while for me to understand the problem! I'm not convinced that another dependency is required to solve this. I'll put some more thoughts in the issue thread.

PR #246, opened 2023-06-05
Add support for Python 3.12

joerick:
~don't merge, just for testing in CI~


joerick:
Struggling with windows test. The github config previously was ignoring errors in the tests. as a result, there is an error in the pstats renderer which wasn't picked up at the time.

Additionally, on 3.12 we have a problem that [importing trio is segfaulting](https://github.com/python-trio/trio/issues/2714), caused by https://foss.heptapod.net/pypy/cffi/-/issues/565#note_304559 .

Will have to skip those tests for now.

PR #254, opened 2023-07-12
Stop built-in code showing as 'None' in traces

### LINKED ISSUE #252: functions like time.sleep are reported as "sleep None"
Follow up from #223 

Profiling the toy code in the issue results in the following output
```
1.019 <module>  profile.py:1
└─ 1.019 A  profile.py:16
   ├─ 0.705 D  profile.py:3
   │  └─ 0.705 sleep  None
   │        [2 frames hidden]  ..
   └─ 0.314 sleep  None
         [2 frames hidden]  ..
```

The line `sleep None` comes from the console renderer [accessing](https://github.com/joerick/pyinstrument/blob/b18e304b560cd9095d46b5bf20b90835e731437e/pyinstrument/renderers/console.py#L184) the [`code_position_short` frame property](https://github.com/joerick/pyinstrument/blob/b18e304b560cd9095d46b5bf20b90835e731437e/pyinstrument/frame.py#L204), which has no fallback in the `if` condition. What would a better repr look like?

joerick:
Fixes #252.

The issue here was a simple coding error, `self.line_no` was `0` so it was falling back as you said @mattip. I decided to keep the if as it was, though, since a line_no of 0 is still invalid, line numbers start at 1, but it will fallback to just the file_path_short, if it's available.

### DIFF
--- a/pyinstrument/frame.py
+++ b/pyinstrument/frame.py
@@ -204,6 +204,7 @@ class Frame:
     def code_position_short(self) -> str | None:
         if self.file_path_short and self.line_no:
             return "%s:%i" % (self.file_path_short, self.line_no)
+        return self.file_path_short
 
     _children: list[Frame]
     attributes: dict[str, float]
--- a/pyinstrument/frame.py
+++ b/pyinstrument/frame.py
@@ -151,24 +151,29 @@ class Frame:
             return self.parent.file_path_short
 
         if not hasattr(self, "_file_path_short"):
+            result = None
+
             if self.file_path:
-                result = None
+                if len(self.file_path.split(os.sep)) == 1:
+                    # probably not a file path at all, more likely <built-in>
+                    # or similar
+                    result = self.file_path
+                else:
+                    for path in sys.path:
+                        # On Windows, if self.file_path and path are on
+                        # different drives, relpath will result in exception,
+                        # because it cannot compute a relpath in this case.
+                        # The root cause is that on Windows, there is no root
+                        # dir like '/' on Linux.
+                        try:
+                            candidate = os.path.relpath(self.file_path, path)
+                        except ValueError:
+                            continue
 
-                for path in sys.path:
-                    # On Windows, if self.file_path and path are on different drives, relpath
-                    # will result in exception, because it cannot compute a relpath in this case.
-                    # The root cause is that on Windows, there is no root dir like '/' on Linux.
-                    try:
-                        candidate = os.path.relpath(self.file_path, path)
-                    except ValueError:
-                        continue
+                        if not result or (len(candidate.split(os.sep)) < len(result.split(os.sep))):
+                            result = candidate
 
-                    if not result or (len(candidate.split(os.sep)) < len(result.split(os.sep))):
-                        result = candidate
-
-                self._file_path_short = result
-            else:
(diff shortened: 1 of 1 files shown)


mattip:
Profiling the code from the issue now shows this, which is better but still has `../<builtin>` which is strange
```
1.020 <module>  profile.py:1
└─ 1.020 A  profile.py:16
   ├─ 0.704 D  profile.py:3
   │  └─ 0.704 sleep  ../<built-in>
   │        [2 frames hidden]  ..
   └─ 0.316 sleep  ../<built-in>
         [2 frames hidden]  ..

```

mattip:
I think `../<built-in>' is coming from the logic in `frame.file_path_short`. It should have a fast-path for `self.file_path == "<built-in>".

joerick:
Weirdly I don't see this `..` issue, how do you invoke pyinstrument to get this?

Though I think your solution sounds right.

mattip:
Using this python on macOS
```
%  python
Python 3.10.10 (v3.10.10:aad5f6a891, Feb  7 2023, 08:47:40) [Clang 13.0.0 (clang-1300.0.29.30)] on darwin
```

I create a venv
```
python -m venv /tmp/venv3
```

then install pyinstrument with `pip install -e .`. I put the script in `/tmp/profile.py` then run
```
% python -m pyinstrument /tmp/profile.py
```

If I stop inside `frame.file_path_short`, I don't see `''` in sys.path, so the shortest `relpath` does not start with `''`, rather with `'..'`

joerick:
Weird.... I get

```console
$ python -m pyinstrument /tmp/profile.py

  _     ._   __/__   _ _  _  _ _/_   Recorded: 19:13:42  Samples:  5
 /_//_/// /_\ / //_// / //_'/ //     Duration: 1.020     CPU time: 0.003
/   _/                      v4.5.0

Program: /tmp/profile.py

1.018 <module>  profile.py:1
└─ 1.018 A  profile.py:16
   ├─ 0.704 D  profile.py:3
   │  └─ 0.704 sleep  <built-in>
   │        [2 frames hidden]  <built-in>
   └─ 0.314 sleep  <built-in>
         [2 frames hidden]  <built-in>

    pyinstrument --load-prev 2023-07-13T19-13-42 [options]
```

I'm on python 3.10.7, on mac.

joerick:
Are you on homebrew python? could that be the difference?

joerick:
Ah, I managed to recreate it, by using a venv in a different directory

```console
$ python -m pyinstrument /tmp/profile.py 

  _     ._   __/__   _ _  _  _ _/_   Recorded: 19:18:10  Samples:  4
 /_//_/// /_\ / //_// / //_'/ //     Duration: 1.019     CPU time: 0.001
/   _/                      v4.5.0

Program: /tmp/profile.py

1.019 <module>  profile.py:1
└─ 1.019 A  profile.py:16
   ├─ 0.703 D  profile.py:3
   │  └─ 0.703 sleep  ../Users/joerick/Projects/pyinstrument/<built-in>
   │        [2 frames hidden]  ..
   └─ 0.315 sleep  ../Users/joerick/Projects/pyinstrument/<built-in>
         [2 frames hidden]  ..

    pyinstrument --load-prev 2023-07-13T19-18-10 [options]
```

Issue #223, opened 2022-11-22
Followup for IPython integration


telamonian:
I got some feedback from the guys I implemented #217 for:

>For:
```
$ cat profiler.py
import time

def D():
    time.sleep(0.7)

def C():
    __tracebackhide__ = True
    time.sleep(0.1)
    D()

def B():
    __tracebackhide__ = True
    time.sleep(0.1)
    C()

def A():
    time.sleep(0.1)
    B()

A()

```
>Profiler output:
```
1.004 <module>  <string>:1
   1.004 run_path  runpy.py:245
      1.001 _run_module_code  runpy.py:89
         1.001 _run_code  runpy.py:63
            1.001 <module>  profiler.py:1
               1.001 A  profiler.py:16
                  0.700 D  profiler.py:3
                     0.700 sleep  None
                        [2 frames hidden]  <built-in>
                  0.301 sleep  None
                     [2 frames hidden]  <built-in>

```
>
>1) Inside `A`, we would like to see that a few frames (B&C) were hidden because of `__tracebackhide__`. We also suggest including a >way to expand hidden nodes in HTML format.
>2) I see, `sleep None`. Should it be `<built-in>` instead? [_Note: I don't see None in Jupyter notebook_].

I'm not entirely sure what of the above is bugs, intended behavior, or new feature requests. The request for expandable hidden nodes in html format seems like a reasonable new feature. I can start looking into making that happen.

As for "Inside `A`, we would like to see that a few frames (B&C) were hidden because of `__tracebackhide__`", I think what they mean is that when a frame is hidden due to the `__tracebackhide__` rule, the output should explicitly say that.

@joerick Can you please clarify exactly what the above example profiling output is currently saying about the hidden frames? I assume the first `[2 frames hidden]  <built-in>` line is refering to the `B` and `C` scopes. Is that right? In that case, the `<built-in>` label is a bug, and should say `<ipython built-in>` (or something) instead. What does the second repeat of the `[2 frames hidden]  <built-in>` refer to?

joerick:
> Can you please clarify exactly what the above example profiling output is currently saying about the hidden frames?

The [2 frames hidden] should mean that there are library frames here (not application frames) that are hidden by default. That functionality is represented in the data model as a 'Group'. But I also agree that there is a bug here... I get the following-

```
pyinstrument /tmp/profilet.py     main

  _     ._   __/__   _ _  _  _ _/_   Recorded: 19:04:16  Samples:  5
[...]

PR #259, opened 2023-08-06
make tracebackhide nodes hidden in html renderer

mattip:
Continuation of #190 and #211, #217, #223. There was a comment about
> We also suggest including a way to expand hidden nodes in HTML format.

that was not addressed.

This implements keeping tracebackhile nodes by:
- removing the `processors.remove_tracebackhide` on the html renderer
- add a check for hiding tracebackhide nodes in the `should_be_hidden` function

This is enough to change the output of the test in #223 

```
python -m pyinstrument -r html -o /tmp/throw_away.html /tmp/test_tracbackhidden.py
```
to (where the "hidden" nodes are expanded)
Screenshot from 2023-08-06 12-40-35

The tui output is unchanged:
```
$ python -m pyinstrument /tmp/test_tracbackhidden.py 

  _     ._   __/__   _ _  _  _ _/_   Recorded: 12:42:10  Samples:  4
 /_//_/// /_\ / //_// / //_'/ //     Duration: 1.002     CPU time: 0.001
/   _/                      v4.5.1

Program: /home/matti/oss/pyinstrument/pyinstrument/__main__.py /tmp/test_tracbackhidden.py

1.001 <module>  test_tracbackhidden.py:1
└─ 1.001 A  test_tracbackhidden.py:16
   ├─ 0.700 D  test_tracbackhidden.py:3
   │  └─ 0.700 sleep  <built-in>
   └─ 0.301 sleep  <built-in>

    pyinstrument --load-prev 2023-08-06T12-42-10 [options]
```

### DIFF
--- a/pyinstrument/processors.py
+++ b/pyinstrument/processors.py
@@ -128,6 +128,8 @@ def group_library_frames_processor(frame: Frame | None, options: ProcessorOption
             return False
         if should_hide:
             return True
+        if frame.has_tracebackhide:
+            return True
 
         return not frame.is_application_code
 
--- a/pyinstrument/renderers/base.py
+++ b/pyinstrument/renderers/base.py
@@ -93,7 +93,8 @@ class FrameRenderer(Renderer):
                 # processors.remove_first_pyinstrument_frames_processor,
                 #    (still hide the outer pyinstrument calling frames)
             ):
-                self.processors.remove(p)
+                if p in self.processors:
+                    self.processors.remove(p)
         if timeline:
             self.processors.remove(processors.aggregate_repeated_calls)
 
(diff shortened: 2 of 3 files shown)


joerick:
i'm kinda torn on this one... it feels like the role of `__traceback_hide__` is to not appear in stack traces, where the code is self-declaring "i'm just a wrapper, I'm not important". That's different from the way that the HTML output collapses frames, which is pyinstrument making an assumption on behalf of the user that they should _probably_ look at their own code before looking in library stack traces. I feel that `__traceback_hide__` is a much stronger signal, so it should be respected (that is, unless --show-all is passed).

Also, I think this might have problems with e.g. decorators where there's a single stack frame that wants to be hidden - this would appear in the pyinstrument tree as a FrameGroup of size 1, as a result, it wouldn't be hidden at all (IIRC).

joerick:
Closing as a wontfix.

PR #258, opened 2023-07-23
Reset sys.argv after running child script/module

joerick:
Fixes #241.

Using `trace` as an example of a script that changes sys.argv:

Before:
```console
$ python -m pyinstrument -m trace --trace wikipedia_article_word_count.py
...
  _     ._   __/__   _ _  _  _ _/_   Recorded: 17:06:48  Samples:  476
 /_//_/// /_\ / //_// / //_'/ //     Duration: 0.819     CPU time: 0.437
/   _/                      v4.5.1

Program: trace

0.818 <module>  trace.py:1
...
```

After:

```console
$ python -m pyinstrument -m trace --trace wikipedia_article_word_count.py
...
  _     ._   __/__   _ _  _  _ _/_   Recorded: 17:08:07  Samples:  482
 /_//_/// /_\ / //_// / //_'/ //     Duration: 0.821     CPU time: 0.444
/   _/                      v4.5.1

Program: /Users/joerick/Projects/pyinstrument/pyinstrument/__main__.py -m trace --trace wikipedia_article_word_count.py

0.821 <module>  trace.py:1
...
```

aside: Note this doesn't change pyinstrument's way of passing sys.argv to children. The behaviour is that all of 

- `python script.py arg1 arg2`
- `python -m script.py arg1 arg2`
- `pyinstrument script.py arg1 arg2`

and

- `python -m module arg1 arg2`
- `python -m pyinstrument -m module arg1 arg2`
- `pyinstrument -m module arg1 arg2`
 
look the same from the child's perspective. That's verified in the test https://github.com/joerick/pyinstrument/blob/af87938d8647676dd185ecf6913ec5c30e225825/test/test_cmdline.py#L81
.

That test was faulty as written, I've also fixed that in this PR.

PR #260, opened 2023-08-07
HTML renderer: show session name in header

matthiasdiener:
The session name is currently only shown in the tab title, as far as I can see. The tab title is often cut off, and can't be easily copied. 

Here is how it looks like:

image


joerick:
this screenshot doesn't look good, it breaks the layout.

perhaps something like this would work better?

Desktop Copy

joerick:
font settings for this are-
image

matthiasdiener:
> this screenshot doesn't look good, it breaks the layout.
> 
> perhaps something like this would work better?
> 
> Desktop Copy

This looks great! Do you have a patch already that shows the session like that, or would you like me to go ahead and create one?

joerick:
i don't this was just something I mocked up in Sketch a while ago

joerick:
so if you're interested in making a patch for this, please do!

matthiasdiener:
Here is what I came up with:

image

joerick:
Cheers @matthiasdiener, I just tweaked it a bit to put the code in App.svelte, I prefer to keep the 'header' component just rendering the header block.

matthiasdiener:
Thanks @joerick ! Is there any chance for a new release? The changes in this PR and in #258 have been very useful!

joerick:
yes, I will do soon, I have been working on 3.12 support but just merged that so I'll do a release very soon. Thanks for the ping :)

PR #261, opened 2023-08-08
Update CI components

matthiasdiener:
- Update versions of some actions components
- Change to Python 3.11 release, add 3.12-dev

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -8,12 +8,12 @@ jobs:
   pre-commit:
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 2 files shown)


joerick:
Thanks @matthiasdiener! I'm working on 3.12 support over in #246, it's complicated by lots of the dev dependencies not supporting 3.12 yet themselves. The trickiest one being cffi, relied on by trio, is segfaulting on import! I think I'll just have to disable those tests.

matthiasdiener:
Ah, I see. Sorry for the spam!

PR #266, opened 2023-09-13
Add method to write directly to html file

thomasaarholt:
Hello! Thanks for writing pyinstrument - I use it frequently, and absolutely love the html representation of the output!

This PR introduces a method `Profiler.write_html` in order to write `output_html` straight to a html file. It replaces the following pattern, which I use frequently:

```python
with Profiler() as profiler:
    # some_slow_code
Path("profile.html").write_text(profiler.output_html())
```

New pattern:

```python
with Profiler() as profiler:
    # some_slow_code
profiler.write_html("profile.html")
```

I'm unsure whether to add a test or not, since I would be basically testing `pathlib.Path.write_text`.

### DIFF
--- a/pyinstrument/profiler.py
+++ b/pyinstrument/profiler.py
@@ -2,10 +2,10 @@ from __future__ import annotations
 
 import inspect
 import os
-from pathlib import Path
 import sys
 import time
 import types
+from pathlib import Path
 from time import process_time
 from typing import IO, Any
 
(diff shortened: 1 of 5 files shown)


pyinstrument/profiler.py
```diff
@@ -304,6 +306,13 @@ def output_html(self, timeline: bool = False) -> str:
         """
         return self.output(renderer=renderers.HTMLRenderer(timeline=timeline))
 
+    def write_html(self, path: str | os.PathLike[str], timeline: bool = False):
+        """
+        Writes the profile output as HTML to a file, as rendered by :class:`HTMLRenderer`
+        """
+        file = Path(path)
+        file.write_text(self.output(renderer=renderers.HTMLRenderer(timeline=timeline)))
```
joerick, line 314: The HTML file actually specifies utf-8 via a `<meta charset="utf-8">` tag, so let's write the file as that. It'll also workaround encoding errors on platforms who have strange encodings.

joerick (commented):
Seems like a good idea, yeah, why not! Just a small request below.

pyinstrument/profiler.py
thomasaarholt: Great point! I should have caught that with my diff! Added!

thomasaarholt:
Great!

Issue #270, opened 2023-10-06
Significant performance overhead when timing requires a syscall on Linux


J08nY:
~Hi, I recently saw a performance regression while using pyinstrument. It seems to be connected to the 4.5.1 -> 4.5.3 update but also might be related to Python version (3.9.17 to 3.9.18).~
Update: This is likely related to sympy library use.

I have two Github CI runs which use pyinstrument to perf some code that uses sympy.

### Good run
https://github.com/J08nY/pyecsca/actions/runs/6039852294/job/16389425413

In the good run, pyinstrument is 4.5.1 and Python is 3.9.17 and the logs look reasonable
(look [here](https://github.com/J08nY/pyecsca/actions/runs/6039852294/job/16389425413#step:10:218) for example, it takes about 1.2 seconds).

### Bad run
https://github.com/J08nY/pyecsca/actions/runs/6247569280/job/16960430761

In the bad run, pyinstrument is 4.5.3 and Python is 3.9.18 and the logs look very long
(look [here](https://github.com/J08nY/pyecsca/actions/runs/6247569280/job/16960430761#step:10:218) for example, it now takes 13.6 seconds). I understand the logs are quite verbose since I have the `show_all` option
enabled, but that does not explain the runtime increase.

### Local test
Note that there are some additional calls to sympy added in commits between the two runs.
However I believe these are not the cause of the performance regression, since I can observe the regression
locally.

Running the code without pyinstrument I get about 5 second runtime:
```python
start = datetime.now()
for _ in range(operations):
    one_point = dbl(p256.curve.prime, one_point, **p256.curve.parameters)[0]
end = datetime.now()
print(end - start)
```
```text
Profiling 5000 256-bit doubling formula (dbl2015rcb) executions...
0:00:05.042373
````

Running with pyinstrument I get about 90 seconds:
```text
Profiling 5000 256-bit doubling formula (dbl2015rcb) executions...

  _     ._   __/__   _ _  _  _ _/_   Recorded: 17:00:13  Samples:  90372
 /_//_/// /_\ / //_// / //_'/ //     Duration: 90.888    CPU time: 90.885
/   _/                      v4.5.3

Program: pyecsca/test/ec/perf_formula.py

90.888 main  test/ec/perf_formula.py:12
└─ 90.859 DoublingEFDFormula.__call__  pyecsca/ec/formula.py:223
   ├─ 85.403 DoublingEFDFormula.__validate_assumptions  pyecsca/ec/formula.py:154
[...]

Issue #83, opened 2019-12-27
pyinstrument is very slow on django tests with docker


patroqueeet:
Dear,

I'm trying to use pyinstrument on docker with a django test setup. When I run the command like this `./manage.py test --keepdb services.rest.tests.test_views:AddCompanyTestCase.test_add_company -v 3` I get something like 
```
240.152 <module>  manage.py:2
└─ 239.407 execute_from_command_line  django/core/management/__init__.py:378
      [23227 frames hidden]  django, django_nose, nose, unittest, ...
         48.646 patched  mock/mock.py:1314
         └─ 48.625 test_add_company  services/rest/tests/test_views.py:59
            ├─ 21.667 post  rest_framework/test.py:296
            │     [770 frames hidden]  rest_framework, django, reversion, se...
            │        21.055 inner  django/core/handlers/exception.py:31
            │        └─ 21.055 __call__  company/middleware.py:28
            │           └─ 21.036 inner  django/core/handlers/exception.py:31
            │              └─ 21.036 __call__  project/middleware.py:10
            │                 └─ 21.036 inner  django/core/handlers/exception.py:31
            │                       [106 frames hidden]  django, rest_framework, re, sre_compi...
            │                          20.995 dispatch  rest_framework/views.py:470
            │                          └─ 20.982 post  services/rest/views.py:813
            │                             └─ 20.851 save  rest_framework/serializers.py:171
            │                                └─ 20.851 create  services/rest/serializers.py:279
            │                                   ├─ 9.561 __get__  django/db/models/fields/related_descriptors.py:373
            │                                   │     [72 frames hidden]  django, raven, weakref
            │                                   ├─ 6.187 now  django/utils/timezone.py:218
            │                                   │     [2 frames hidden]  django
            │                                   └─ 4.289 [self]
            ├─ 19.563 reverse  django/urls/base.py:27
            │     [718 frames hidden]  django, re, sre_compile, sre_parse, e...
            │        4.988 import_module  importlib/__init__.py:109
            │        └─ 4.972 <module>  project/urls.py:1
            └─ 7.285 generate  project/generators.py:151
               └─ 7.281 _make_user  project/generators.py:64
                  ├─ 3.613 _make_username  project/generators.py:54
                  │  └─ 3.612 _make_wordlist  project/generators.py:47
                  │     └─ 3.612 <listcomp>  project/generators.py:49
                  │        └─ 3.602 [self]
[...]

PR #273, opened 2023-10-14
Reduce timing overhead

joerick:
Addressing #270.

Couple of approaches.

If CLOCK_MONOTONIC_COARSE is available on the system and the desired profiling interval is slow enough, that timer is used. This timer is very low overhead, as it just accesses a variable in user-space, and never makes a syscall.

Alternatively, if the user requests it, the profiler will run a timing thread. This is a C thread that runs outside the GIL and updates a variable with the current time every \<interval\> seconds.


joerick:
I think I've figured out the ctypes thing. It seems that loading the `.so` file with the path works on both linux and mac.

dralley:
Also potentially addressing https://github.com/joerick/pyinstrument/issues/83

pyinstrument/low_level/stat_profile.c
```diff
@@ -141,9 +99,12 @@ static double ProfilerState_GetTime(ProfilerState *self) {
...
         Py_DECREF(result);
         return resultDouble;
+    } else if (self->timer_thread_subscription_id >= 0) {
+        // when a self->timer_thread_subscription_id is set, use the timing thread.
+        return pyi_timing_thread_get_time();
     } else {
-        // otherwise as normal, call the C timer function.
-        return floatclock();
+        // otherwise as normal, call the synchronous C timer function.
+        return pyi_floatclock(self->interval);
```
joerick, line 107: Need to pass the float clock type here

PR #278, opened 2023-10-27
Disable variable expansion in IPython magic

### LINKED ISSUE #277: Interpolation bug in jupyter magic
Hi,

We came across a bug in the jupyter magic (https://github.com/joerick/pyinstrument/blob/main/pyinstrument/magic/magic.py):
```
>>> x = 1
>>> %pyinstrument print("$x")
1
```
This should print "$x" and not interpolates it. The same is true for things in `{}` like `%pyinstrument print("hello {len('world')}")` (`%time print("hello {len('world')}")` works as expected).

While I'm not sure why this happens, I think moving to the now officially supported way to do this [ast_mod](https://ipython.readthedocs.io/en/stable/api/generated/IPython.core.magics.ast_mod.html) would fix it (at the very least a simple test shows ast_mod does not have this bug).

Would it be ok to make a PR to move to this, knowing it only works for IPython 8.15+ ([release notes](https://ipython.readthedocs.io/en/stable/whatsnew/version8.html#ipython-8-15))? If not, maybe we can do a version check and newer IPython versions use this and older code falls back to what we have today as _most_ users don't seem to have run into this.

krassowski:
Fixes #277.

Uses [`@no_var_expand`](https://ipython.readthedocs.io/en/stable/api/generated/IPython.core.magic.html#IPython.core.magic.no_var_expand) to disable IPython variable interpolation.

This decorator is supported since in IPython 7.3 and is used in other projects and magics which offer execution of user code in special contexts, including the IPython built-in magics such as `%debug`, `%%prun`, `%time`, or in rpy2's `%R` and `%%R` magics.

### DIFF
--- a/pyinstrument/magic/magic.py
+++ b/pyinstrument/magic/magic.py
@@ -4,7 +4,7 @@ import urllib.parse
 from ast import parse
 
 from IPython import get_ipython  # type: ignore
-from IPython.core.magic import Magics, line_cell_magic, magics_class
+from IPython.core.magic import Magics, line_cell_magic, magics_class, no_var_expand
 from IPython.core.magic_arguments import argument, magic_arguments, parse_argstring
 from IPython.display import IFrame, display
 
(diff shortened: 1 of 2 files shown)


krassowski:
Just a gentle ping in case if it slipped through the notifications. Please do let me know if anything else is needed here :)

joerick:
thanks for the gentle ping :) looks good to me.

PR #285, opened 2023-12-15
Add option to show total time including children in flat profiles

matt-graham:
Adds a literal string argument `flat_time` to `ConsoleRenderer` which can be used to set frame time type displayed (and used as sort key) in flat profiles - `"self"` for total self time, excluding child frames (the default, equivalent to current behaviour) and `"total"` for total time, including child frames. 

I find being able to view a flat list sorted by total time useful in some situations, but it does result in the less intutitive behaviour that the sum of the times shown no longer bears any relation to total run time. 

Rather than a string argument could alternatively use a boolean flag if preferred or change name of options.

### DIFF
--- a/pyinstrument/renderers/console.py
+++ b/pyinstrument/renderers/console.py
@@ -26,6 +26,7 @@ class ConsoleRenderer(FrameRenderer):
         color: bool = False,
         flat: bool = False,
         time: LiteralStr["seconds", "percent_of_total"] = "seconds",
+        flat_time: LiteralStr["self", "total"] = "self",
         **kwargs: Any,
     ) -> None:
         """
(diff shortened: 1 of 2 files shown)


matt-graham:
For the remaining `pyright` error in `pre-commit` workflow, I get the same error when running `pre-commit run --all-files` locally on current tip of `main` so not sure if the failure is related to changes in this PR or a change in upstream packages?

joerick (commented):
This looks fine to me, as-is.  I don't actually use the flat renderer, but if it's useful to you, all good!

Could you add the option to `Profiler.print` and `Profiler.output_text`, too?

matt-graham:
> Could you add the option to `Profiler.print` and `Profiler.output_text`, too?

`flat_time` option now also exposed in `Profiler.print` and `Profiler.output_text`. I created a type alias `FlatTimeMode` to avoid repetition of the allowable string modes in the type hint and this seemed to mirror the pattern used for `AsyncMode` in `profiler.py` but let me know if you prefer this to be named differently.

joerick (approved):
Looks good, thanks!

PR #294, opened 2024-03-05
Add flat argument to print to Profiler.print()

gnuille:
Currently flat can only be enabled via command line:
```
pyinstrument -p flat myscript.py
```
Add the flat argument to the `Profiler.print` method so it can be used via the python API.


dpinol:
Hi,
what do you think about this?
thanks

PR #301, opened 2024-04-17
Ensure Profiler convenience methods expose render options

paugier:
Thank you for pyinstrument, which is really nice, useful and easy to use!

In order to print percentages, I had to change a `profiler.print()` to

```python
print(
    profiler.output(
        renderer=ConsoleRenderer(
            unicode=True,
            color=True,
            show_all=False,
            time="percent_of_total",
         )
    )
)
```

With these changes, I think it would have been enough to write `profiler.print(time="percent_of_total")`.

This PR is to explain the situation and discuss what would be a good solution.


joerick:
yeah, ideally we'd do this, the issue is that the [documentation](https://pyinstrument.readthedocs.io/en/latest/reference.html#pyinstrument.Profiler.print) is generated from the code signature, and having kwargs breaks that link. Perhaps a better way would be to write a unit test that checks that the signatures of `print` and `output_text` match that of `ConsoleRenderer.__init__`.

joerick:
Or maybe there's a way to get sphinx to pull the option definitions through? It would be fine to link to the descriptions, but I'd like the options themselves listed, like [`textwrap.wrap`](https://docs.python.org/3/library/textwrap.html#textwrap.wrap), for example.

PR #305, opened 2024-05-28
Use slightly more recent IPython API.

Carreau:
The new ast transformer are a bit more flexible instead of just appending/prepending some code and allow full Python constructs.

- you can use ``__code__`` for where the original code should go, and ``__ret__``
- by default it can also mangle all names that start with `___` (3 underscore), if you need some more hygienic templating. The names are still leaked in global namespace but with invalid python indentifiers names starting with `mangle-` (where `-` is invalid in an identifier).

This allow us to later clean the namespace manually in the magic.

And this also allwo us to also return the original value from the cell, I don't know if this is something desirable, but I added it nonetheless.

### DIFF
--- a/pyinstrument/magic/magic.py
+++ b/pyinstrument/magic/magic.py
@@ -2,14 +2,15 @@ import asyncio
 import threading
 import urllib.parse
 from ast import parse
+from textwrap import dedent
 
+import IPython
 from IPython import get_ipython  # type: ignore
 from IPython.core.magic import Magics, line_cell_magic, magics_class, no_var_expand
 from IPython.core.magic_arguments import argument, magic_arguments, parse_argstring
 from IPython.display import IFrame, display
 
 from .. import Profiler
-from ._utils import PrePostAstTransformer
 
 _active_profiler = None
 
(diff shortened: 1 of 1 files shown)


Carreau:
Note that i'm writing that as part of trying to investigate a case where it look like the output of the profiler's IPython magic does show all the frame above the root frame where the profiler is started:

```
35.438 MainThread  <thread>:139648662820672
`- 35.437 <module>  ../../../../../../bin/ipython:1
   `- 35.437 start_ipython  IPython/__init__.py:104
      `- 35.437 TerminalIPythonApp.launch_instance  traitlets/config/application.py:1069
         `- 35.437 TerminalIPythonApp.start  IPython/terminal/ipapp.py:320
            `- 35.437 TerminalInteractiveShell.mainloop  IPython/terminal/interactiveshell.py:906
               `- 35.437 TerminalInteractiveShell.interact  IPython/terminal/interactiveshell.py:890
                  `- 35.437 TerminalInteractiveShell.run_cell  IPython/core/interactiveshell.py:3083
                     `- 35.437 TerminalInteractiveShell._run_cell  IPython/core/interactiveshell.py:3125
                        `- 35.437 _pseudo_sync_runner  IPython/core/async_helpers.py:120
                           `- 35.437 TerminalInteractiveShell.run_cell_async  IPython/core/interactiveshell.py:3224
                              `- 35.437 TerminalInteractiveShell.run_ast_nodes  IPython/core/interactiveshell.py:3459
                                 `- 35.437 <module>  <ipython-input-3-70d0375aba7c>:1
                                    `- 35.437 TerminalInteractiveShell.run_line_magic  IPython/core/interactiveshell.py:2443
                                       `- 35.437 PyinstrumentMagic.pyinstrument  pyinstrument/magic/magic.py:44
                                          `- 35.437 TerminalInteractiveShell.run_cell  IPython/core/interactiveshell.py:3083
....

```

But I do not have access to a reproducer. So I'm mostly trying to understand the code, see where this could coming from and suggesting improvements.

joerick:
Thanks. BTW, the part that is supposed to trim that off is `Session._trim_stem()`. https://github.com/joerick/pyinstrument/blob/main/pyinstrument/session.py#L132-L150

Carreau:
> Thanks. BTW, the part that is supposed to trim that off is `Session._trim_stem()`. https://github.com/joerick/pyinstrument/blob/main/pyinstrument/session.py#L132-L150

Thanks that is helpful.

Carreau:
Ok, let's mark that ready for review. I was not able to fix the bug yet and will do in a separate pr.

Carreau:
Thanks !

PR #309, opened 2024-06-10
Allow non-vendored deps to be used

keszybz:
Distributions will want to get rid of the vendored dependencies and use the system versions. Modify the imports to try the bundled copy first, and then fall back to the system one. This keeps behaviour unchanged by default, but allows easy de-vendoring by simply removing the vendored copy from source.

### DIFF
--- a/pyinstrument/__main__.py
+++ b/pyinstrument/__main__.py
@@ -20,7 +20,16 @@ from pyinstrument.util import (
     file_supports_unicode,
     object_with_import_path,
 )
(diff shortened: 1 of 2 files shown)


joerick:
I'm not sure that the files in `vendor` really map to pypi packages - this is probably the only place you could get them. For example, the keypath file in `vendor` contains modifications from the original.

https://github.com/joerick/pyinstrument/blob/a7ba94c0fe0a4c79da051227ac4f8e404d5039d7/pyinstrument/vendor/keypath.py#L33

Issue #318, opened 2024-07-12
Pyinstrument CLI hides script exit code


frantic:
It's very tempting to use `pyinstrument` in production systems for some long-running scripts by just replacing `python <script>` with `pyinstrument <script>`. However, looks like `pyinstrument` hides the exit code of the script it runs.

Consider this example:

```sh
$ cat test.py
import sys
sys.exit(1)

$ python test.py || echo failed
failed

$ pyinstrument test.py || echo failed
  _     ._   __/__   _ _  _  _ _/_   Recorded: 17:42:06  Samples:  1
 /_//_/// /_\ / //_// / //_'/ //     Duration: 0.001     CPU time: 0.001
/   _/                      v4.5.0

Program: test.py

0.001 BufferedReader.read  None

To view this report with different options, run:
    pyinstrument --load-prev 2024-07-11T17-42-06 [options]

```

Note that running `test.py` via `pyinstrument` changed its exit code from `1` to `0`.

The root cause seems to be [here](https://github.com/joerick/pyinstrument/blob/b364e53ce1feaf279bd3082c7002c674ca06c0ac/pyinstrument/__main__.py#L367)

```python
try:
    sys.argv[:] = argv
    exec(code, globs, None)
except (SystemExit, KeyboardInterrupt):
    pass
finally:
    sys.argv[:] = old_argv
```

I think a reasonable fix is to remember the exit code in that exception handler and then `sys.exit` with that code to make sure the caller of the script is aware of the failure.

joerick:
agreed, looks like a bug

PR #319, opened 2024-07-12
Propagate `SystemExit` and `KeyboardInterrupt` from CLI

frantic:
My naive attempt at fixing #318


joerick:
This looks good to me. I actually don't fully know what the effect are of re-raising an exception outside of the except block. I'll have a quick play to see...

joerick:
Works great! thanks

PR #325, opened 2024-08-01
Alt-click to recursive expand in html renderer

### LINKED ISSUE #320: Feature Request: Render stack as collapsed in HTML

joerick:
Fix #320

- Upgrade JS deps


html_renderer/src/lib/model/Frame.ts
```diff
@@ -19,6 +20,7 @@ export default class Frame {
   group: Group | null;
 
   constructor(jsonObject: FrameData, parent: Frame|null = null, context: FrameContext = {groups:{}}) {
+    this.uuid = crypto.randomUUID()
```
Carreau, line 23: this breaks for me. I get 

```
Uncaught TypeError: crypto.randomUUID is not a function
at data:text/html
```

in the embeded script (chrome, latest). Even if `crypto.randomUUID()` from the console does return a UUID. Should crypto be somehow imported ?

joerick: yeah, thats a bug, thanks for reporting. see #328

Carreau: Might just be some caching locally (ignore for now).

joerick: No, i think you're right. This api only works in [secure contexts](https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID). Some users will load pyinstrument from insecure contexts, like http servers. I guess data URIs are classed as insecure contexts too, maybe again the rules are a bit opaque. I've pushed a fix for that anyway.

Issue #324, opened 2024-08-01
Support more options through IPython magic 


singharpit94:
Currently, %%pyinstrument only supports following options through IPython magic:

--interval INTERVAL
--async_mode ASYNC_MODE
--height HEIGHT
--timeline TIMELINE

Can we add support for options like outfile, hide-regex etc ideally all options supported by pyinstrument

PR #334, opened 2024-08-20
Draft: Share option definition between cli and magic. 

Carreau:
There is some request in #324 to have the same options (when possible) between the magic and the CLI. 

I'm looking into defining those options only once, and refactoring main/magic to not repeat the options handling. 

I'm tryin to be the least invasive as possible.

I'm not yet looking for deep review, but mostly letting you know I'm working on this, and maybe cursory validation that the approach is ok (or not), or wether you prefer a more thourough refactor, or simply me redefining the options twice in the `__main__` and the magic.

### DIFF
--- a/pyinstrument/__main__.py
+++ b/pyinstrument/__main__.py
@@ -24,58 +24,6 @@ from pyinstrument.vendor import appdirs, keypath
 
 # pyright: strict
 
-from collections import namedtuple
-
-Option = namedtuple(
-    "Option", "short, long, dest, action, metavar, default, help, callback, type".split(", ")
-)
-
-
-class ParserAdapter:
-    _options: list
-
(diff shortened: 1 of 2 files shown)


Carreau:
I also want to note that pyinsturment is currently using Otparse which is marked as deprecated since 3.2 in favor of argparse. I guess there might be a reason not to use Argparse.

Carreau:
I'm thinking we might want to have and explicit opt-in list instead of opt-out list, and that many of those options like load and co, only make sens with a line magic, not a cell magic.

Carreau:
Ok, i'l see if I can find a better way to do that, and pull some shared logic into `compute_render_options`, maybe there is a way to have shared options defined as data somewhere.

Thanks for the feedback.

PR #339, opened 2024-09-19
update: Adds options to customize profiler middleware saved file file…

HazemBZ:
Gives more control over Django middleware saved filenames


joerick:
Thanks!

PR #350, opened 2024-11-20
Share some options between the CLI and magic.

Carreau:
This refactor compute_render_options, to be shared between CLI and magic.

Add a few of the CLI options to be available in the magic.

The only thing I don't really like is that we redefine the help text options.

### DIFF
--- a/pyinstrument/__main__.py
+++ b/pyinstrument/__main__.py
@@ -409,9 +409,31 @@ def main():
         raise inner_exception
 
 
+class OptionsParseError(Exception):
+    pass
+
+
 def compute_render_options(
-    options: CommandLineOptions, renderer_class: type[renderers.Renderer], output_file: TextIO
+    options: CommandLineOptions,
+    renderer_class: type[renderers.Renderer],
+    unicode_support: bool,
+    color_support: bool,
 ) -> dict[str, Any]:
+    """
+    Given a list of `CommandLineOptions`, compute the
+    rendering options for the given renderer.
+
+    Raises an `OptionsParseError` if there is an error parsing the options.
+
+    unicode_support:
+        indicate whether the expected output supports unicode
+    color_support:
+        indicate whether the expected output supports color
+
+    Both of these will be used to determine the  default of outputting unicode
+    or color, but can be overridden with `options.color` and `option.unicode`.
+    """
+
     # parse show/hide options
     if options.hide_fnmatch is not None and options.hide_regex is not None:
         raise OptionsParseError("You can‘t specify both --hide and --hide-regex")
@@ -449,8 +471,8 @@ def compute_render_options(
     if issubclass(renderer_class, renderers.ConsoleRenderer):
         unicode_override = options.unicode is not None
         color_override = options.color is not None
-        unicode: Any = options.unicode if unicode_override else file_supports_unicode(output_file)
-        color: Any = options.color if color_override else file_supports_color(output_file)
+        unicode: Any = options.unicode if unicode_override else unicode_support
+        color: Any = options.color if color_override else color_support
 
         render_options.update({"unicode": unicode, "color": color})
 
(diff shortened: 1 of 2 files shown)


Carreau:
See discussion in #334

Carreau:
Closing and reopening to trigger CI.

pyinstrument/__main__.py
```diff
@@ -412,9 +412,41 @@ def store_and_consume_remaining(
 def compute_render_options(
     options: CommandLineOptions, renderer_class: type[renderers.Renderer], output_file: TextIO
 ) -> dict[str, Any]:
+    """
+    Given a list of CommandLineOoptionsCompute the rendering options compute the
```
joerick, line 416:
```suggestion
    Given a list of CommandLineOptions, compute the
```

pyinstrument/__main__.py
```diff
@@ -412,9 +412,41 @@ def store_and_consume_remaining(
...
+    renderer_class: type[renderers.Renderer],
+    unicode_support: bool,
+    color_support: bool,
+) -> Tuple[Optional[str], Optional[dict[str, Any]]]:
+    """
+    Similar as compute_render_options, but return a tuple (error message, data)
+    if there is an error; this will let us reuse _compute_render_options in magics.
+
+    output_file, has been replaced by unicode_support:bool, color_support:bool
+    """
```
joerick, line 446: I think we don't need this extra wrapper function, rather, instead we could wrap the invocations of `compute_render_options` in magic.py with `try:...catch:`...? unless i'm missing something?

joerick (commented):
sorry for the delay in the review. it looks good! just a couple comments

Carreau:
> sorry for the delay in the review. it looks good

No needs to apologize, this is open source and users are not owed anything. 
I've learn myself to instead of apologizing, thanking contributor for their patience – at least I try. It's a subtle difference, but at least help me with framing my guilt and decrease burnout. 

I will look back into this when possible, 

Happy new year as well.

pyinstrument/__main__.py
Carreau: I just wanted to avoid contaminating the magic system with `OptionsParseError`, which is defined in `__main__`, and I don't want cyclic imports; but I can make `compute_render_option` not raise as it is used in a single place, and raise `OptionsParseError` in the sole place it it used.

Carreau: Also `compute_render_options`, takes an `output_file` while the private variant takes two booleans (unicode_support: bool, color_support: bool), as there are no output file when running the magic. I can lift both of those as well

joerick: Ah, the unicode/color support bit makes sense. Let's change the `compute_render_options` to use `unicode_support` and `color_support` (moving the file-related stuff to the call site) and then remove the `_compute_render_options` private version. 

I'd rather do the ideomatic thing and raise exceptions on errors. I don't see a problem with `magic.py` importing OptionsParseError, it's already importing from `__main__` anyway. But if it is an issue, we could make it a ValueError and catch that instead.

joerick (commented):
Happy new year! And thanks for your patience. Very wise words, I'll try to take them on ;)

I think a small simplification and this is ready to go. If it's not clear I'm happy to jump in and make the change, just let me know.

Carreau:
> And thanks for your patience

😄, thanks for yours. 

> Ah, the unicode/color support bit makes sense. Let's change the compute_render_options to use unicode_support and color_support (moving the file-related stuff to the call site) and then remove the _compute_render_options private version.

> I'd rather do the idiomatic thing and raise exceptions on errors. I don't see a problem with magic.py importing OptionsParseError, it's already importing from __main__ anyway. But if it is an issue, we could make it a ValueError and catch that instead.

It should be done, I now just let OptionsParseError bubble up, that seem to work if it's ok with you, so I don't even have to import it.

joerick (approved):

joerick:
Great! Looks good.

Issue #352, opened 2024-12-04
Support multithreaded profiling


georgeharker:
Thanks for pyintrument - it's incredibly useful

I had need to trace a multithreaded python app and examine the relationship between threads.  Obviously in some cases multithreading can be a little interesting in python, but in this particular case works well.

I have extended a fork of pyinstrument to support showing all child threads from the one where profiling starts.  I get nice results with all threads separated which has been hugely helpful.

I'll file a PR and reference this issue.  It's still a bit of a WIP, but i'd be curious to see if it looks reasonable to you.

lucamuscat:
This would definitely be a great feature to have in pyinstrument.

As a workaround until/if this feature gets picked up, I've come up with the following hack to automatically profile newly spawned threads by (ab)using `threading.settrace`, [here's an implementation of this hack in one of my projects.]( https://github.com/lucamuscat/perfsephone/pull/17/files#diff-ce33e10775b88333d87d09709caabab4687ab7f9284f7d94f63e354d9230decaR46-R87)

One can detect when a thread has started, and start the profiler when `threading.settrace` calls the tracing function with `Threading.run()` in its `frame` argument with the `call` event. The profiler can be stopped when the same `Threading.run()` is encountered in the trace function with a `return` event. You need to keep track of the number of times `Threading.run()` is called, as a thread can spawn other threads.

georgeharker:
Take a look at the referenced PR - it does much of that but also allows viewing the timelines of all threads together.

lucamuscat:
@georgeharker I've already had a look, I need to dedicate some time on integrating it with my project.

My biggest issue with the PR in its current state is that the majority of tests are failing, there are some missing updates to the Session class in various methods, and the linters seem to be complaining.

Changing the return type for the Profiler from a list of frames to a dictionary seems like a tough breaking change. Maybe this multi threading aware code can be placed into its own class, leaving us with two kinds of profilers?

georgeharker:
Yeah there's still work to do there for sure! Just did enough for my needs and wanted some feedback about what it would look like to get it into a state for merge

PR #353, opened 2024-12-04
Support multithreaded profiling

georgeharker:
As per #352 I have modified pyinstrument to support multiple threads.

It does this by separating out both timing and frames for the different threads - as such it modifies a substantial amount of the data model (usually in the form of `List[frames]` -> `Dict[str, List[Frames]]` and the equivalents in typescript.

I don't do much typescript so it's possible there are prettier ways of doing the modifications on that side.

- I've modified most of(but not all) of the renderers (I don't use / have speedscope so I'm not sure of the best fix there).
- I'm sure the ux could be tidied, the threads come out in arbitrary order in the timeline view and it would ideally be possible to easily toggle various threads off etc).

I'd be interested in discussing how this looks to you


lucamuscat:
Hey :wave: 

Good work!

It seems that you still need to insert the `thread_start_times` field in `Session.from_json` & `Session.combine`

pyinstrument/stack_sampler.py
```diff
@@ -24,7 +25,15 @@
...
 
-StackSamplerSubscriberTarget = Callable[[List[str], float, Optional["AsyncState"]], None]
+SubscriberCallstackFn = Callable[[List[str], float, Optional["AsyncState"]], None]
+SubscriberEventFn = Callable[[str, str, float], None]
+
+
+@dataclass
+class StackSamplerSubscriberTarget:
+    call_stack: SubscriberCallstackFn
+    event: SubscriberEventFn
```
joerick, line 35: I wonder if we can combine this event and async_state change concepts into one interface. It seems that they're pretty similar to me.

pyinstrument/profiler.py
```diff
@@ -30,13 +34,21 @@ def __init__(
...
         self.frame_records = []
         self.target_description = target_description
-        self.interval = interval
+        self.interval = interval,
+        self.child_threads = child_threads
+
+    def record_thread_start(self, thread_id: str, time: float) -> None:
+        if not self.thread_start_times:
+            self.first_start_time = time
+        self.thread_start_times[thread_id] = time - self.first_start_time
```
joerick, line 51: Why not store the raw start time in each, rather than the offset from the `first_start_time`? Generally I find it better to store truth over computed values.

pyinstrument/profiler.py
```diff
@@ -246,6 +269,19 @@ def __enter__(self):
...
+            time: float):
+        if not self._active_session:
+            raise RuntimeError(
+                "Received a state change without an active session. Please file an issue on pyinstrument Github describing how you made this happen!"
+            )
+        if event == 'thread_start':
+            self._active_session.record_thread_start(thread_id, time)
+
     # pylint: disable=W0613
     def _sampler_saw_call_stack(
```
joerick, line 286: I'm confused why this method doesn't need to change? Don't we need to separate out the storage of the different threads?

joerick (commented):
Thanks for sending this over. It's certainly ambitious! But it would be a good improvement to get this working. I'm still a bit confused about how this works. Do you have a sample script that illustrates how it's supposed to work?

pyinstrument/profiler.py
georgeharker: I initially tried raw time but it had some complexity when trying to view the traces on the same view, i'd be happy to look at that again.

pyinstrument/profiler.py
georgeharker: The frame records get a root thread id added which allows us to reconstruct what is separate - this is done in build_call_stack

georgeharker: (different threads will never clash in the pythons code due to the GIL)

georgeharker:
Done!

> Hey 👋
> 
> Good work!
> 
> It seems that you still need to insert the `thread_start_times` field in `Session.from_json` & `Session.combine`

Done!

lucamuscat:
> Hi @lucamuscat 😄 
> 
> 
> 
> Is there anything stopping this being merged?

Hey 👋
I'm not a maintainer of this project, I had just left some comments as I was also looking at this PR and was hoping to help out in getting it merged.

@joerick is the right person to ask.

georgeharker:
There were some unresolved comments that I can help with (though a little snowed under work wise). I don't think it's super far off

PR #356, opened 2024-12-19
Export renderers from `pyinstrument.renderers`

rmartin16:
Type checkers don't like it when you use symbols that are not exported in that module. Putting the renderers in `__all__` will export them here so imports like this one pass mypy:

```
from pyinstrument.renderers import HTMLRenderer
```

Currently one must specify the specific module that owns the renderer to import it.

```
from pyinstrument.renderers.html import HTMLRenderer
```


joerick (approved):
thanks!

PR #373, opened 2025-05-06
Update profiler.py: fix mypy valid-type error

wyattscarpenter:
Mypy, a typechecker, complains when code using this library uses the exposed type AsyncMode, for reasons explained here https://github.com/python/mypy/issues/8674 and here https://mypy.readthedocs.io/en/stable/common_issues.html#variables-vs-type-aliases — this patch fixes that problem by adding a TypeAlias type annotation.

I'm not convinced this is valuable behavior on mypy's part, but I am using mypy, and pyinstrument, so, might as well comply with its wishes, to shut it up 🙂

### DIFF
--- a/pyinstrument/profiler.py
+++ b/pyinstrument/profiler.py
@@ -7,20 +7,15 @@ import time
 import types
 from pathlib import Path
 from time import process_time
-from typing import TYPE_CHECKING, IO, Any
+from typing import IO, Any
 
 from pyinstrument import renderers
 from pyinstrument.frame import AWAIT_FRAME_IDENTIFIER, OUT_OF_CONTEXT_FRAME_IDENTIFIER
 from pyinstrument.renderers.console import FlatTimeMode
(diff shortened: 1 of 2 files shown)


wyattscarpenter:
What an interesting pattern of failures... typing_extensions is only installed in 3.11 and before, I suppose... 🤔 there are multiple easy ways this could be fixed.

wyattscarpenter:
Thanks for mentioning! That narrows it down considerably! 🙂 I'll get right on that...

wyattscarpenter:
Alright, the only failure left now is
```
Run pre-commit/action@v2.0.3
install pre-commit
Error: getCacheEntry failed: This legacy service is shutting down, effective April 15, 2025. Migrate to the new service ASAP. For more information: https://gh.io/gha-cache-sunset
```
Which I surmise is some unrelated problem, and the TYPE_CHECKING workaround worked.

Issue #388, opened 2025-07-25
Time duration formatting for better readability


Cycloctane:
Frame time durations are always displayed in seconds, which can be ambiguous and even loss precision when a frame is several milliseconds or below 1 millisecond. It would be nice if console and html renderer can adaptively format time duration into more readable units based on its value and interval time setting.

e.g.
`0.068` -> `68ms` (default interval)
`0.003` -> `3ms` (when default interval) -> `3ms 360μs` (when interval=10μs)

PR #390, opened 2025-07-31
Adaptive frame time display precision

Cycloctane:
Format frame time base on precision of session interval time to improve readability and accuracy. (#388 )


joerick (commented):
Really nice work on this @Cycloctane! I appreciate the typescript fixes - normally I have to make those myself!

Issue #341, opened 2024-10-04
Customize profile filename in Django middleware


maingoh:
It would be nice to easily be able to customize the profile filename in the middleware https://github.com/joerick/pyinstrument/blob/4b37f8cdc531be41a7f7e57932f0b770244025d5/pyinstrument/middleware.py#L78

maingoh:
Looks related to #339

PR #393, opened 2025-08-06
Update/django middleware custom filname callback

HazemBZ:
Addresses https://github.com/joerick/pyinstrument/issues/341


joerick (approved):
Thanks!

Issue #401, opened 2025-09-30
Missing preprocessors and their options for HTMLRenderer


martinm37:
Hello, since 5.0.0 there seem to be missing `__init__` arguments for the `preprocessors` and `processor_options` for the `HTMLRenderer`. Because of this, the instance of `JSONForHTMLRenderer` that `HTMLRenderer` creates cannot have these attributes, and because of that there is no way to shorten the output `.html` file. This is particularly an issue when the output file has 600MB or more when profiling a particularly large block of code.

Could these options be added back into the `HTMLRenderer`?

`HTMLRenderer` is able to have these options if I add these arguments to the `__init__` method:

``` 
def __init__(
        ...
        preprocessors: ProcessorList = [],
        processor_options: dict[str, Any] | None = None,
    ):
        ...
        self.preprocessors: ProcessorList = preprocessors
        self.processor_options: dict[str, Any] | None = processor_options
```
and when I add this to the `render` method:

```
    def render(self, session: Session):
        ...
        json_renderer.processor_options = self.processor_options
        ...
```

joerick:
Yeah, we can add these back, I think. The reason the processors were removed is that they were reimplemented in the web browser to be used at runtime.

However, it turned out I could save a lot of Typescript code by sending an _unprocessed_ frame-tree rather than a list of samples. This was considered an implementation detail at the time, but I can see that being able to reduce the size of sample files is a decent use-case.

If you'd like to add a PR for this I think it can be fairly easy to accept.

PR #403, opened 2025-10-11
fix(HTMLRenderer): re-implementing processors and their options

martinm37:
I have done 3 changes:

1) I have added back the `processors` and `processor_options` attributes into the `HTMLRenderer` class, so that `JSONForHTMLRenderer` class can have access to them, when created by `HTMLRenderer`.

2) In the process of doing 1), I have renamed attribute `HTMLRenderer.preprocessors` to `HTMLRenderer.processors`, as it seems more logical. Because of this `magic.py` had to be updated.

3) I have added these 2 attributes into the `Profiler.output_html`, as without them one test was not passing.

I have also wanted to update the online docs, but I did not find a way how to change the `docs/reference.md`.

This PR addresses the issue #401

### DIFF
--- a/docs/reference.md
+++ b/docs/reference.md
@@ -93,6 +93,7 @@ Therefore, rendering can be customised by changing the ``processors`` property. 
 .. autoclass:: pyinstrument.renderers.ConsoleRenderer
 
 .. autoclass:: pyinstrument.renderers.HTMLRenderer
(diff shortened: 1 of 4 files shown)


martinm37:
Hello, I have tried it out now, and it works as it should, i.e. I am able hide the function calls which take small amount of time, and thus reduce the output file size. Thank you very much once again!!!

PR #423, opened 2026-01-29
Support multiple -o/--outfile flags in a single invocation

### LINKED ISSUE #422: Support multiple -o/--outfile flags in a single invocation
It would be useful to generate multiple output formats in one run

e.g.: pyinstrument -o output.txt -o output.html script.py

Use case: I like to get the HTML view for visual inspection and the text view to pass to an LLM, since it's fewer tokens. Currently I work around this by outputting a .pyisession file and then rendering it to HTML and text separately its a small step

Behavior:
- Each outfile infers its renderer from the file extension (unless -r is explicitly set, which applies to all)
- No -o or single -o behavior is unchanged
- Unrecognized extensions without -r produce a clear error

ryan-crabbe:
## Summary
- Change `-o`/`--outfile` to accept multiple values via `action="append"`, so users can write multiple output formats in one invocation
- Each outfile infers its renderer from the file extension (unless `-r` is explicitly set, which applies to all)
- No `-o` or single `-o` behavior is unchanged
- Adds tests for multiple outfiles, single outfile, explicit renderer with multiple outfiles, and unrecognized extension error

Closes #422

### DIFF
--- a/docs/guide.md
+++ b/docs/guide.md
@@ -27,6 +27,11 @@ Here are the options you can use:
 **Protip:** `-r html` will give you a interactive profile report as HTML - you
 can really explore this way!
 
+You can also specify `-o` multiple times to generate multiple output formats in
+a single run. Each outfile infers its renderer from the file extension:
+
+    pyinstrument -o profile.txt -o profile.html script.py
+
 ## Profile a Python CLI command
 
 For profiling an installed Python script via the
--- a/docs/reference.md
+++ b/docs/reference.md
@@ -12,6 +12,21 @@ print a profile report to the console.
 ```{program-output} pyinstrument --help
 ```
 
+### Output file (`-o`/`--outfile`)
+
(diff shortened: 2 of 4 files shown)


joerick:
I like the idea, but I don't like the change to existing behaviour - previously the -r option took precidence over the file extension to determine format. 

Perhaps one way to fix this would be to also allow multiple `-r` commands, so that,

- if there's one, we use the current behaviour (all outfiles use that renderer)
- if there's more than one, each `-r` affects the following `-o outfile` arguments.
- when no `-r` is passed, we use the file extension.

The other (maybe more punative way) would be to just disallow `-r` when using multiple outfiles. This might be preferable if the code complexity to acheive the above is too high.

It would also be good to make sure we're documenting the `outfile` extension inferring behaviour in the `--help` output - as we increasingly use file extension to determine format, we need to make sure the docs and helptext explain this (and make the various formats discoverable).

PR #446, opened 2026-07-13
fix(magic): use ConsoleRenderer for text config instead of HTMLRenderer

### LINKED ISSUE #419: Jupyter plugin ignores --render-option=time=percent_of_total
Or maybe I'm doing something wrong? I'm trying to have the Jupyter magic automatically render percentages instead of seconds:

Cell 1:

```python
%load_ext pyinstrument
import numpy as np

def create_ones_array():
    return np.ones((1_000_000,), dtype=np.int64)

def do_some_math():
    array = create_ones_array()
    for i in range(1000):
        array2 = create_ones_array()
        array += array2 * array.sum()
    return array
```

Cell 2:

```python
%%pyinstrument --render-option=time=percent_of_total
_ = do_some_math()
```

The `percent_of_total` is ignored. In normal Jupyter I'm getting percentages by default, in Quarto (a publishing system that uses Jupyter under the hood) I'm getting seconds, in either case I can't control the output.

PyInstrument is version 5.1.1.

Also I can't figure out how to get it to have the timeline be selected by default when generating the profile in Jupyter.

fuleinist:
## Summary

The Jupyter `%%pyinstrument` magic calls `compute_render_options` twice — once for the HTML renderer config and once for the text renderer config. Both calls were passing `renderer_class=HTMLRenderer`, which means renderer-specific options for `ConsoleRenderer` (like `unicode`, `color`, and `time`) were never set in the text config.

This caused `--render-option=time=percent_of_total` (and other `ConsoleRenderer`-specific options) to be silently ignored in the text output.

## Fix

Changed the second call to use `renderer_class=ConsoleRenderer` so that:

1. **`unicode` and `color`** are properly set in the text config (they were previously being hardcoded to `True` at line 196–197, which masked the bug for those specific options)
2. **`time`** and other `ConsoleRenderer`-specific constructor kwargs (like `flat`, `flat_time`, `short_mode`) are now correctly parsed from `--render-option` flags
3. **`processor_options`** (hide/show regex) are properly passed to the text renderer

## Testing

- `--render-option=time=percent_of_total` now correctly shows percentages in the text output
- `--hide`, `--show`, `--show-all` continue to work for both HTML and text output
- `--unicode`, `--color` flags continue to work for both renderers

Fixes #419


joerick (approved):

joerick:
thanks!

PR #452, opened 2026-07-25
fix: respect PYINSTRUMENT_SHOW_CALLBACK when PYINSTRUMENT_PROFILE_DIR is set

### LINKED ISSUE #342: PYINSTRUMENT_SHOW_CALLBACK should not be ignored if PYINSTRUMENT_PROFILE_DIR is defined
I defined `PYINSTRUMENT_PROFILE_DIR` to record multiple requests, but I also would like to ignore some of them to avoid having too much reports being generated. So I would to be able to set `PYINSTRUMENT_PROFILE_DIR` and also `PYINSTRUMENT_SHOW_CALLBACK`. But according the the code, when `PYINSTRUMENT_PROFILE_DIR` is defined, all requests are recorded.

Thus I think we could remove the `or` condition here https://github.com/joerick/pyinstrument/blob/4b37f8cdc531be41a7f7e57932f0b770244025d5/pyinstrument/middleware.py#L52 as by default `PYINSTRUMENT_SHOW_CALLBACK = True`.

Sanjays2402:
Closes #342

The Django middleware's condition was `(show_pyinstrument(request) and url_argument in request.GET) or profile_dir`, so as soon as `PYINSTRUMENT_PROFILE_DIR` was set every request was profiled and `PYINSTRUMENT_SHOW_CALLBACK` was ignored. The callback now gates both branches, so it can be used to select which requests get written to the profile directory. Regression test fails on main (`assert not hasattr(request, "profiler")`) and passes with the fix.

### DIFF
--- a/docs/guide.md
+++ b/docs/guide.md
@@ -179,6 +179,9 @@ def custom_show_pyinstrument(request):
 PYINSTRUMENT_SHOW_CALLBACK = "%s.custom_show_pyinstrument" % __name__
 ```
 
+This callback also applies when `PYINSTRUMENT_PROFILE_DIR` is set, so it can be
+used to select which requests get written to the profile directory.
+
 You can configure the profile output type using setting's variable `PYINSTRUMENT_PROFILE_DIR_RENDERER`.
 Default value is `pyinstrument.renderers.HTMLRenderer`. The supported renderers are
 `pyinstrument.renderers.JSONRenderer`, `pyinstrument.renderers.HTMLRenderer`,
--- a/pyinstrument/middleware.py
+++ b/pyinstrument/middleware.py
@@ -46,10 +46,10 @@ class ProfilerMiddleware(MiddlewareMixin):  # type: ignore
         else:
             show_pyinstrument = lambda request: True
 
-        if (
-            show_pyinstrument(request)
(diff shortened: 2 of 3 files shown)


joerick (approved):
Looks good. Thanks. 

I really like how the test is written. If you had time to make a few more tests like this for Django and check it was all run in CI that would be great!

joerick:
+1 to the copilot comment and once the formatter is happy we can get it in.

Sanjays2402:
Both done in a2e224d.

Switched the two tests to pytest's `tmp_path` (passed to `override_settings` as `str(tmp_path)`) instead of the hard-coded `/tmp/pyinstrument-test-profiles`, so they're isolated per-test and work on Windows.

The formatter complaint was black wanting the `process_request` condition on one line rather than split across two — ran black and isort over the touched files.

```
$ python -m pytest test/test_django_middleware.py -q
..                                                                       [100%]
2 passed in 3.34s

$ black --check test/test_django_middleware.py pyinstrument/middleware.py
All done! 2 files would be left unchanged.
```

Note `test_cmdline.py` and `test_ipython_magic.py` are also flagged by the current black version, but they're untouched by this PR so I left them alone.

joerick:
thanks!

